PR #223, opened 2015-04-11
.gitignore file with some improvements

GMLudo:
Hi,

This is the .gitignore equivalent of your .hgignore.
I've improved a little bit, I've used the Python template for .gitignore.
I've tested with `setup.py sdist` and `setup.py sdist`, no issues.

Thanks for the merge.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,75 @@
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+
+# C extensions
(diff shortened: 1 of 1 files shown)


asvetlov:
I've just converted .hgignore and .hgeol to git equivalents in 1888b1d
Thank you for reporting an issue.

GMLudo:
Ok, thank you, no problems with that ;-)

PR #229, opened 2015-04-11
Rename asyncio.async to asyncio.create_task

1st1:
Since we're considering adding async/await to python at some point, I think we should rename 'async' function to 'create_task', since it does essentially the same thing.

asyncio has a provisional status in standard library, so while this is a breaking change we can do it.

### DIFF
--- a/asyncio/base_events.py
+++ b/asyncio/base_events.py
@@ -293,7 +293,7 @@ class BaseEventLoop(events.AbstractEventLoop):
         self._check_closed()
 
         new_task = not isinstance(future, futures.Future)
-        future = tasks.async(future, loop=self)
+        future = tasks.create_task(future, loop=self)
         if new_task:
             # An exception is raised if the future didn't complete, so there
             # is no need to log the "destroy pending task" message
--- a/asyncio/tasks.py
+++ b/asyncio/tasks.py
@@ -2,7 +2,7 @@
 
 __all__ = ['Task',
            'FIRST_COMPLETED', 'FIRST_EXCEPTION', 'ALL_COMPLETED',
-           'wait', 'wait_for', 'as_completed', 'sleep', 'async',
+           'wait', 'wait_for', 'as_completed', 'sleep', 'create_task',
            'gather', 'shield',
            ]
 
@@ -327,7 +327,7 @@ def wait(fs, *, loop=None, timeout=None, return_when=ALL_COMPLETED):
     if loop is None:
         loop = events.get_event_loop()
 
-    fs = {async(f, loop=loop) for f in set(fs)}
+    fs = {create_task(f, loop=loop) for f in set(fs)}
 
     return (yield from _wait(fs, timeout, return_when, loop))
 
@@ -361,7 +361,7 @@ def wait_for(fut, timeout, *, loop=None):
     timeout_handle = loop.call_later(timeout, _release_waiter, waiter)
     cb = functools.partial(_release_waiter, waiter)
 
-    fut = async(fut, loop=loop)
+    fut = create_task(fut, loop=loop)
     fut.add_done_callback(cb)
 
     try:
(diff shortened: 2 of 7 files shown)


asvetlov:
Please no. 
Don't remove existing function from asyncio without keeping trampoline for old code (with raising DeprecationWarning of course).

kxepal:
So you want to break API and create two different asyncio branches for 3.3 and 3.4 versions? Please, don't.

1st1:
@kxepal we will push an update to 3.4 too.

It's an easy thing to "unbreak" in your code for all possible python/asyncio versions:

     try:
         from asyncio import create_task
     except ImportError:
         from asyncio import async as create_task

kxepal:
@1st1 In fact, in my, and I guess, in most of code there is only `asyncio.async` call, not `async` because of oblivious reasons. So this trick wouldn't work.

More over, wouldn't such name create a confusion with `loop.create_task`?

1st1:
Well, if you have access to the event loop object you should use `loop.create_task`. `asyncio.async` is just a handy wrapper on top of it. Since the functionality of these two methods is essentially the same, I think that the proposed change will actually streamline the API and help avoid users having any confusion with what they should use.

gvanrossum:
No, async() is not just a wrapper on top of create_task() -- it only creates a task if the argument is not already a Future. I recommend naming it ensure_task(). We can keep the old name as well.

haypo:
> Since we're considering adding async/await to python at some point, I
think we should rename 'async' function to 'create_task', since it does
essentially the same thing.

Really? Is there a PEP? Is it accepted? IMO yield from is enough.

No need to modify asyncio *right now*.

1st1:
@haypo The PEP is in the making, and yes, we'll wait a bit to see how it goes.

"async" is still a bad name for this function. So we'll have figure out a way to keep it in asyncio anyways (but just deprecated).

haypo:
Renaming a very important function is just wrong. What about backward compatibility.

This issue doesn't solve any real issue, I just prefer to close it. Come back when the PEP is accepted.

GMLudo:
[joke] As a simple end-user of AsyncIO, I have a strong opinion on that, because for one time, I understand what you're talking about ;-) [/joke]

More seriously:

1. First, I'm strongly agree with @haypo: I don't know for you, but at least for me, we have more than 40+ daemons on production based on this version of AsyncIO API.
Yes, we don't upgrade all daemons in the same time, nevertheless, even if this change is pretty simple to handle, I estimate that we have pay enough the price to be an early adopter on AsyncIO: We invest time to understand AsyncIO, ecosystem and improve it for our needs. Please respect your simple end-users: change only to add features/bugfixes, or improve performances. For a better end-user experience with API, it should be a new version.
2. Second, nevertheless, I think that a positive enhancement should be taken from @1st1 remark for the next main release: In AsyncIO, you have several global functions like `asyncio.async`, `asyncio.as_completed`, `asyncio.gather`, `asyncio.sleep` where you have an optional `loop` parameter. If you don't give a `loop` parameter, you use the implicit global event loop defined.
But, for some others features, like `create_server` or `create_datagram_endpoint`, there are methods of loop, the developer must take more explicitly the decision to use global event loop defined via `asyncio.get_event_loop()` or use a `loop` variable.
Moreover, it should be easier for newcorners that all functions that use a loop would be methods of loop object, instead of to check the documentation each time to remember if it's a loop method or asyncio function.
I understand your goal, especially for `asyncio.sleep`: to be quicker to use in a coroutine, you don't need to retrieve `loop`. Maybe you made that because you've also some technicals constraints.
Personally, I can live with this actual API, it's only a suggestion.

I can help to write a PEP and/or implement that, it seems to be enough simple for me, but before to do something, what's your opinion ? (Don't forget I wasn't implied in AsyncIO ecosystem day 1, if you have some pointers on discussions you could already have about that, I'm interested in).

ambv:
First of all, @haypo, `asyncio` is provisional so talking about backwards compatibility is simply invalid at this point. However, this is *the time* to rename `@asyncio.async` if we ever consider adding `async/await`. By the time we release Python 3.5, it's going to be too late.

Second, @gvanrossum, if we leave the old name intact, this effectively blocks the `async/await` PEP.

@haypo, @gvanrossum - I don't expect you to blindly accept this without reading the actual proposal but maybe let's not close this before we actually rejected async/await as an idea?

1st1:
@ambv @GMLudo Guys, let's wait a little bit on this. I'm cooking the PEP right now.

In any case, this pull request will be heavily augmented to keep old 'async' function (although it will be deprecated) in a way that is compatible with our ideas about async/await.

Also, there might be a way to add async/await without future import at all and without breaking backwards compatibility. I'm investigating this right now ;)

GMLudo:
@1st1 ok, np, I'm waiting, good cooking ;-)

Issue #222, opened 2015-04-10
Debug mode assumes coroutine callables have __name__


GoogleCodeExporter:
On python 3.4.3, the following code will work when running without debug mode 
enabled, and raise an exception when run with PYTHONASYNCIODEBUG=1:
```
import asyncio
import functools

@asyncio.coroutine
def func(x, y):
    print('func with x = {} and y = {}'.format(x, y))
    yield from asyncio.sleep(0.1)
    print('exit func')


partial_func = asyncio.coroutine(functools.partial(func, 1))

loop = asyncio.get_event_loop()
try:
    loop.run_until_complete(partial_func(2))
finally:
    loop.close()
```

Line 154 of asyncio.coroutines tries to copy some attributes of the wrapped 
function to its CoroWrapper instance, which fails because the partial object 
has no `__name__`:
```
AttributeError: 'functools.partial' object has no attribute '__name__'
```

Original issue reported on code.google.com by `mwfrojd...@gmail.com` on 31 Mar 2015 at 2:16

haypo:
I wrote a patch to fix the attribute error, but there is still an issue: the program behaves differently in debug mode.

gvanrossum:
When I run the program with `PYTHONASYNCIODEBUG=1`, with the fix in place, I get a traceback:
```
<CoroWrapper func() running at p.py:4, created at /Users/guido/src/asyncio/asyncio/coroutines.py:141> was never yielded from
Coroutine object created at (most recent call last):
  File "p.py", line 15, in <module>
    loop.run_until_complete(partial_func(2))
```
(etc.) but I think this is a red herring, or at least an unrelated issue -- you shouldn't be using partial() to wrap a coroutine, because you're really wrapping the CoroWrapper object at that point. Removing the `@asyncio.coroutine` decorator in front of `def func` resolves this. I've filed #241 about this (but there I also explain I think it's a bug in the example program).

PR #231, opened 2015-04-14
Fix @coroutine for functions without __name__

haypo:
Issue #222: Fix the @coroutine decorator for functions without __name__
attribute like functools.partial().

Enhance also the representation of a CoroWrapper if the coroutine
function is a functools.partial().

### DIFF
--- a/asyncio/coroutines.py
+++ b/asyncio/coroutines.py
@@ -151,7 +151,8 @@ def coroutine(func):
             w = CoroWrapper(coro(*args, **kwds), func)
             if w._source_traceback:
                 del w._source_traceback[-1]
-            w.__name__ = func.__name__
+            if hasattr(func, '__name__'):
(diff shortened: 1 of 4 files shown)


gvanrossum:
LGTM (I'm wary of the automatic merge, on Bitbucket I ended up with lots of dummy merge commits due to that).

kxepal:
@gvanrossum if you want to avoid merge commits, you can do the following trick:
1. Add access to PRs for your git cli: https://gist.github.com/piscisaureus/3342247
2. Rebase master against PR
3. Push

GitHub will automatically close PR as merged since it will see that PR commit hashes are on master branch. But this only works if PR is already rebased against origin master. Otherwise merge commit is imminent.

gvanrossum:
Just doing the web merge.

PR #237, opened 2015-04-20
Fix queue join tests for CPython's test runner.

ajdavis:
"python -m test" was running tests in the base class, which failed.


gvanrossum:
Thanks, the test runner issue is fixed. I'll merge this to 3.4 and 3.5. I still get the error in the log I reported, but it appears to be a timing issue -- sometimes it's there, sometimes not. And I can't see how it would be your fault given where it occurs.

ajdavis:
Thanks! I looked over my code and the tests and I agree about the "pending Task" warning, it must have preceded my changes.

PR #239, opened 2015-04-23
Add option to cancel all children on first raised exception to gather

### LINKED ISSUE #238: Gather: Optionally cancel all children on first raised exception
I'm running two separate coroutines on a websockets connection by using ``asyncio.gather``. One sends keep alive messages (*ping*) and the other one handles messages on the connection. If the *ping* times out, the connection should be closed. And if the connection closes, sending keep alive messages is pointless. I'm raising an exception when the *ping* times out and I thought that this would cancel the other task as well. Turns out, it doesn't.

Now, here comes the problem: Since ``asyncio.gather`` handles the coroutines in self-created tasks, there is no way to cancel them any more from the outside (well, at least not without accessing private members). Cancelling the returned future will not work either because the future is marked as ``done``.

I propose a keyword argument that could be named ``exceptions_cancel_tasks`` for ``asyncio.gather``. I'll create a pull request shortly.

lgrahl:
This adds another optional argument ``exceptions_cancel_tasks``. If set to ``True`` the first raised exception (that is not a ``CancelledError``) in a child will cancel any other children.

Closes #238

### DIFF
--- a/asyncio/tasks.py
+++ b/asyncio/tasks.py
@@ -536,7 +536,8 @@ class _GatheringFuture(futures.Future):
         return True
 
 
-def gather(*coros_or_futures, loop=None, return_exceptions=False):
+def gather(*coros_or_futures, loop=None, return_exceptions=False,
+           exceptions_cancel_tasks=False):
     """Return a future aggregating results from the given coroutines
     or futures.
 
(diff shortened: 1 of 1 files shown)


gvanrossum:
Hm, I don't think I like this. gather() is getting too many flags this way, and they can't even be combined. I also don't think this is common (I don't recall being asked for this before) and going by the description you gave in #238 there's an easy workaround. Not every three-line function deserves to be enshrined in a library.

lgrahl:
I have to say, I am not a big fan of the mutual exclusive flags either. I had another idea to add a *force* flag to the cancel function of ``_GatheringFuture`` or add another function that cancels the remaining children. But this would mean that ``gather()`` doesn't return a regular ``Future`` instance any more which certainly is overkill for an uncommon problem. In the end, I think you and @asvetlov are correct - there is an easy workaround for this kind of problem.

PR #242, opened 2015-05-11
Deprecate async() function in favour of ensure_future()

1st1:
This is something that PEP 492 requires us to do -- rename the 'async()' function. Guido suggested to use 'ensure_future()' name instead.

### DIFF
--- a/asyncio/tasks.py
+++ b/asyncio/tasks.py
@@ -3,7 +3,7 @@
 __all__ = ['Task',
            'FIRST_COMPLETED', 'FIRST_EXCEPTION', 'ALL_COMPLETED',
            'wait', 'wait_for', 'as_completed', 'sleep', 'async',
-           'gather', 'shield',
+           'gather', 'shield', 'ensure_future',
            ]
 
 import concurrent.futures
@@ -12,6 +12,7 @@ import inspect
 import linecache
 import sys
 import traceback
+import warnings
 import weakref
 
 from . import coroutines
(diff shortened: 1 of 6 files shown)


asyncio/tasks.py
```diff
@@ -500,6 +501,20 @@ def async(coro_or_future, *, loop=None):
     """Wrap a coroutine in a future.
 
     If the argument is a Future, it is returned directly.
+
+    This function is deprecated in 3.5. Use asyncio.ensure_future() instead.
+    """
+
+    warnings.warn("asyncio.async() function is deprecated, use ensure_future()",
```
gvanrossum: I wonder if we should put off warnings until at least one asyncio release with both functions has been out. asyncio releases follow Python 3.4.x release.

gvanrossum:
LGTM, please merge by hand and comment out the warning for now. (Or make it a silent deprecation with a comment to make it heavier in the next release.)

1st1:
Would `DeprecationWarning` work?  It's silent.

gvanrossum:
Perfect. (But after 1-2 releases I'd like to make the warning non-silent.)

On Mon, May 11, 2015 at 10:10 AM, Yury Selivanov <notifications@github.com>
wrote:

> Would DeprecationWarning work? It's silent.
>
> —
> <https://github.com/python/asyncio/pull/242#issuecomment-100981176>.
>



-- 

1st1:
This PR has been merged.  Thanks for the review, Guido!

Issue #233, opened 2015-04-19
Accept generator created in Cython


methane:
Problem was reported here: https://groups.google.com/forum/#!topic/cython-users/g146SZHxRyM
I think supporting Cython is very important for asyncio performance.

gvanrossum:
Agreed, I would gladly accept patches. I'm not a Cython user myself so I can't fix this without help.

GMLudo:
Hi @methane, thank you for the link.
On the mailing-list, Stefan seems to find a solution: https://github.com/cython/cython/commit/4af42443bd37f6207143f8870904f780a65bb3e3
I've played my role of messenger: I've invited him to open a pull request.

scoder:
I had already created a CPython ticket for this:

https://bugs.python.org/issue24004

Here's what Cython currently does in order to circumvent the type checks in asyncio and inspect:

https://github.com/cython/cython/blob/4af42443bd37f6207143f8870904f780a65bb3e3/Cython/Utility/Generator.c#L824

https://github.com/cython/cython/blob/4af42443bd37f6207143f8870904f780a65bb3e3/Cython/Utility/Generator.c#L906

It's not clear to me what the best fix is here, but my gut feeling tells me that the correct way to deal with this is to create a Generator ABC and do the type checking based on that. A "Generator" is an object that implements the Generator protocol, not necessarily something of type `types.GeneratorType`. Special casing that type is fine, obviously, but excluding everything else isn't.

The fact that this crash bug here went unnoticed since Py3.3 makes it obvious, though, that no-one has ever tried this at the C level:

https://bugs.python.org/issue23996

methane:
I think adding `Coroutine` ABC to somewhere is good way to solve this problem.
It may be in the stdlib.  But I think we should wait discussion of async/await syntax.

For now, how about `asyncio.coroutines.Coroutine` ?

```python
from abc import ABCMeta, abstractmethod
import types

class Coroutine(metaclass=ABCMeta):
    def __iter__(self):
        return self

    def __next__(self):
        raise StopIteration

    @abstractmethod
    def send(self):
        raise StopIteration


Coroutine.register(types.GeneratorType)
Coroutine.register(types.CoroWrapper)


def iscoroutine(obj):
    """Return True if obj is a coroutine object."""
    return isinstance(obj, Coroutine)
```

1st1:
[...]

PR #245, opened 2015-05-12
Use collections.abc.Coroutine for asyncio.iscoroutine() when available.

1st1:
This should help to resolve issues like #233.

Should I also include 'collections.abc.Generator'?


asvetlov:
Not every generator is a coroutine.

1st1:
> Not every generator is a coroutine.

That was the original request in #233. However, the newly added `collections.abc.Coroutine` should be enough.

asvetlov:
LGTM

1st1:
This PR was merged in 3a09a93277afc2cdb43badf92a2c85c2789813f6.

PR #246, opened 2015-05-13
Removed needless extra `try` in `run_forever`

Lucretiel:


gvanrossum:
I'm sorry, I don't find that better or clearer. It's only the _run_once() call that can raise _StopError.

1st1:
@gvanrossum OTOH it's only `while True: _run_once()` code in that `try` block...

PR #252, opened 2015-07-07
Fix assert expression

mayflaver:
Fix assert expression


gvanrossum:
Disagree. The error is that you shouldn't pass a Handle since this function expects a callback (i.e. a function/callable). Your change doesn't help understand that.

mayflaver:
A callback is  a Handle？

asvetlov:
`callback` is a regular callable (function, method etc), neither `asyncio.Handle` nor coroutine instance.

Issue #248, opened 2015-05-21
Add transport.closing property


asvetlov:
In aiohttp project I have an issue (KeepSafe/aiohttp#370).

Long story short: for handling static files aiohttp uses code like:

            with open(filepath, 'rb') as f:
                chunk = f.read(self.limit)
                while chunk:
                    resp.write(chunk)
                    yield from resp.drain()
                    chunk = f.read(self.limit)

When client closes HTTP connection (by socket shutdown for example) `transport._force_close(...)` schedules `_call_connection_lost` on next loop iteration and assigns `transport._closing` flags (as well as `transport._conn_lost`).

Actual transport closing will be done on next loop iteration, but aiohttp static file handler has no chance to be informed about transport closing: `stream.write()` and underlying `transport.write()` don't check for `._closing` flag and always succeed. 
`transport.write()` does check for `._conn_lost` but sends log message only, not raises exception -- and the behavior is pretty correct.

aiohttp static file handler sends no data on `resp.write(chunk)` call, stream buffer is never overloaded and `yield from resp.drain()` always returns without pausing.
Thus all multi-megabyte file may be iterated over and pushed into stream. Actual data will not be sent via wire of course but the whole process is

a) takes longer than required
b) sends a lot of warning messages `'socket.send() raised exception.'` to asyncio logger

I propose to add public readonly bool property `.closing` (return `._closing` internal value).
It gives me a way to check if transport is on the middle of closing procedure.
`aiohttp.StreamWriter` also should be modified: I guess to add `.closing` property to `StreamWriter` to mimic transport. 
`StreamWriter.drain()` coroutine should check for `stream.closing` first and call `yield from asyncio.sleep(0)` to run previously scheduled `transport._call_connection_lost` at first.

Sorry for long message, I hope I've described my problem well enough. Feel free to ask if my (complex enough) scenario is still unclean.

If no objections I will prepare a patch.

GMLudo:
+1
I don't understand everything, but I trust in you to push a good solution in aiohttp ;-)

1st1:
> I propose to add public readonly bool property .closed (return ._closed internal value).
> It gives me a way to check if transport is on the middle of closing procedure.

The fact that '.closed' will be set while the transport is in the process of *being* closed (not *already* closed) bothers me a bit, but I think I'm OK with that.

Would be great to hear what @gvanrossum and @haypo think about this first.

gvanrossum:
[...]

PR #257, opened 2015-07-20
Issue #248: Add closing read-only property to transports

haypo:
* Disallow write() on closing transports
* Disallow aslo calling pause_writing() and resume_writing() on StreamReaderProtocol if the transport is closing

### DIFF
--- a/asyncio/streams.py
+++ b/asyncio/streams.py
@@ -153,15 +153,25 @@ class FlowControlMixin(protocols.Protocol):
         self._paused = False
         self._drain_waiter = None
         self._connection_lost = False
+        self._transport = None
+
+    def connection_made(self, transport):
+        self._transport = transport
 
     def pause_writing(self):
         assert not self._paused
+        if self._transport is not None and self._transport.closing:
+            raise RuntimeError('Cannot call pause_writing() '
+                               'on closing or closed transport')
         self._paused = True
         if self._loop.get_debug():
             logger.debug("%r pauses writing", self)
 
     def resume_writing(self):
         assert self._paused
+        if self._transport is not None and self._transport.closing:
+            raise RuntimeError('Cannot call resume_writing() '
+                               'on closing or closed transport')
         self._paused = False
         if self._loop.get_debug():
             logger.debug("%r resumes writing", self)
@@ -217,6 +227,7 @@ class StreamReaderProtocol(FlowControlMixin, protocols.Protocol):
         self._client_connected_cb = client_connected_cb
 
     def connection_made(self, transport):
+        super().connection_made(transport)
         self._stream_reader.set_transport(transport)
         if self._client_connected_cb is not None:
             self._stream_writer = StreamWriter(transport, self,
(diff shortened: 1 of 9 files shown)


asvetlov:
LGTM

asyncio/streams.py
```diff
@@ -153,15 +153,25 @@ def __init__(self, loop=None):
         self._paused = False
         self._drain_waiter = None
         self._connection_lost = False
+        self._transport = None
+
+    def connection_made(self, transport):
+        self._transport = transport
```
gvanrossum: I think you should call `super().connection_made(transport)` here.

tests/test_selector_events.py
```diff
@@ -1000,8 +1000,8 @@ def test_write_closing(self):
         transport = self.socket_transport()
         transport.close()
         self.assertEqual(transport._conn_lost, 1)
-        transport.write(b'data')
-        self.assertEqual(transport._conn_lost, 2)
+        # write() is disallowed after close()
```
gvanrossum: This worries me greatly. This test was pretty clearly testing that calling write() after close() is allowed but just bumps the `_conn_lost` counter.

asyncio/streams.py
```diff
@@ -153,15 +153,25 @@ def __init__(self, loop=None):
...
 
     def pause_writing(self):
         assert not self._paused
```
gvanrossum: Perhaps change these asserts to RuntimeErrors too?

gvanrossum:
So honestly this worries me, so please don't commit yet. FWIW I do think it's fine if closing remains set after the transport is *closed*.

gvanrossum:
I really don't like that `write()` can raise an exception before `connection_lost()` has been called. It feels like a violation of the state machine specification. But exposing the `closing` flag is fine. Can you try a patch with just that?

1st1:
I was hit by this too -- I need to implement `transport._closing` in uvloop.  It's a private API, but it's used a lot in asyncio and in third-party code.  Can we add a simple method `Transport.is_closing()` for this?

gvanrossum:
Can you two agree in whether it should be spelled t.closing or t.closing()
or t.is_closing()? I have no objection against adding it. Make sure to
update the spec of the Transport class.

1st1:
`Transport.is_closing()` is more consistent with `Loop.is_running()` and `Loop.is_closed()`.

And with a quick regex search `def \w+ing\(` I couldn't find methods like `.something()`.

gvanrossum:
Sounds fine.

asvetlov:
Sorry for offtopic but would you, guys, take a look on http://bugs.python.org/issue25074
Approving or rejecting of proposed patch would be great.

1st1:
> Sorry for offtopic but would you, guys, take a look on http://bugs.python.org/issue25074
Approving or rejecting of proposed patch would be great.

If nobody looks at it before Monday I can review it.

gvanrossum:
So, I think we need a new PR that *just * adds is_closing() to the
Transport base class and to all its implementation subclasses. Who wants to
do that? We have until Friday to get it into 3.5.1.

1st1:
Please see PR #291. Closing this one.

PR #259, opened 2015-08-05
Make sure BaseException is re-raised in SSLProtocol

1st1:


gvanrossum:
Is there a bug describing the need for this fix?

1st1:
There is no bug. I was reviewing ``sslproto.py`` code while investigating how to implement start_tls, and this particular spot seemed wrong to my eyes. If the execution is stuck somewhere in ``_process_write_backlog`` (and it's a lot of code involved there), it will take two SystemExits or KeyboardInterrupts to about the execution of the loop.

gvanrossum:
OK, LGTM.

1st1:
Committed in f57cfc623ad.

Issue #270, opened 2015-09-26
Does asycio.Semaphore have the same bug as asyncio.Queue?


gvanrossum:
Check out the last paragraph of https://github.com/python/asyncio/pull/269#issuecomment-143466942:

"""
Looking over the locks code, I also noticed that async.Semaphore (which can be thought of as a queue without data) uses the approach I sketched above (waking up only one waiter at a time), but without reawakening another waiter when a waiter is cancelled after its future was marked done. So I expect it can be shown to have the same bug.
"""

All we need to do now is write a test that reproduces this issue.

gvanrossum:
The scenario would be something like this: have two tasks both blocked in acquire(); have one release() call; then, before the event loop runs the waiting task that was woken up, cancel it; the other waiting task is then just hanging even though the semaphore's semantics would allow it to proceed.

(In the Queue class we've struggled with this and similar scenarios in the opposite direction; the symptoms could be lost items, deadlock, or (in 3.5.0) assertions failing.

manipopopo:
I try to address the issue in #271 .

gvanrossum:
Closed by #271.

PR #271, opened 2015-09-27
Rewrite Semaphore.acquire/release

manipopopo:
For #270

Add a failing test for original Semaphore: when a ready waiter be cancelled, the other pending waiters will not aware of it.

* `test_acquire` in tests/test_locks.py assumes that if *task A* calls `semaphore.acquire()` before *task B*, *A* will acquire the semaphore before *B*. I use `self._num_ready` to make sure that if there is any awoken waiter, the new `acquire` will not cut in line.
* `test_acquire_cancel` in tests/test_locks.py doesn't allow `Semaphore` to leaving cancelled waiters in `Semaphore._waiters`.

I'm not sure whether or not these two tests should be modified.

### DIFF
--- a/asyncio/locks.py
+++ b/asyncio/locks.py
@@ -397,6 +397,7 @@ class Semaphore(_ContextManagerMixin):
         if value < 0:
             raise ValueError("Semaphore initial value must be >= 0")
         self._value = value
+        self._num_ready = 0
         self._waiters = collections.deque()
         if loop is not None:
             self._loop = loop
@@ -410,6 +411,14 @@ class Semaphore(_ContextManagerMixin):
         if self._waiters:
             extra = '{},waiters:{}'.format(extra, len(self._waiters))
         return '<{} [{}]>'.format(res[1:-1], extra)
+
+    def _wake_up_next(self):
+        while self._waiters:
+            waiter = self._waiters.popleft()
+            if not waiter.done():
+                waiter.set_result(None)
+                self._num_ready += 1
+                return
 
     def locked(self):
         """Returns True if semaphore can not be acquired immediately."""
@@ -425,7 +434,7 @@ class Semaphore(_ContextManagerMixin):
         called release() to make it larger than 0, and then return
         True.
         """
-        if not self._waiters and self._value > 0:
+        if not self._num_ready and self._value > 0:
             self._value -= 1
             return True
 
@@ -435,8 +444,16 @@ class Semaphore(_ContextManagerMixin):
             yield from fut
             self._value -= 1
             return True
+        except futures.CancelledError:
+            if self._value > 0:
+                self._wake_up_next()
+            # `test_acquire_cancel` in tests/test_locks.py doesn't allow
+            # cancelled waiters left in self._waiters
+            if fut in self._waiters:
+                self._waiters.remove(fut)
+            raise
         finally:
-            self._waiters.remove(fut)
+            self._num_ready -= 1
 
     def release(self):
         """Release a semaphore, incrementing the internal counter by one.
@@ -444,10 +461,7 @@ class Semaphore(_ContextManagerMixin):
         become larger than zero again, wake up that coroutine.
         """
         self._value += 1
-        for waiter in self._waiters:
-            if not waiter.done():
-                waiter.set_result(True)
-                break
+        self._wake_up_next()
 
 
 class BoundedSemaphore(Semaphore):
--- a/asyncio/locks.py
+++ b/asyncio/locks.py
@@ -443,17 +443,17 @@ class Semaphore(_ContextManagerMixin):
         try:
             yield from fut
             self._value -= 1
+            self._num_ready -= 1
             return True
         except futures.CancelledError:
-            if self._value > 0:
+            if not fut.cancelled():
+                self._num_ready -= 1
                 self._wake_up_next()
             # `test_acquire_cancel` in tests/test_locks.py doesn't allow
             # cancelled waiters left in self._waiters
             if fut in self._waiters:
                 self._waiters.remove(fut)
             raise
-        finally:
-            self._num_ready -= 1
 
     def release(self):
         """Release a semaphore, incrementing the internal counter by one.
--- a/asyncio/locks.py
+++ b/asyncio/locks.py
@@ -432,12 +432,13 @@ class Semaphore(_ContextManagerMixin):
         called release() to make it larger than 0, and then return
         True.
         """
-        while self._value == 0:
+        while self._value <= 0:
             fut = futures.Future(loop=self._loop)
             self._waiters.append(fut)
             try:
                 yield from fut
             except:
+                # See the similar code in Queue.get.
                 fut.cancel()
                 if self._value > 0 and not fut.cancelled():
                     self._wake_up_next()
--- a/asyncio/locks.py
+++ b/asyncio/locks.py
@@ -434,7 +434,7 @@ class Semaphore(_ContextManagerMixin):
         called release() to make it larger than 0, and then return
         True.
         """
-        if not self._num_ready and self._value > 0:
+        if self._value - self._num_ready > 0:
             self._value -= 1
             return True
 
--- a/asyncio/locks.py
+++ b/asyncio/locks.py
@@ -397,7 +397,6 @@ class Semaphore(_ContextManagerMixin):
         if value < 0:
             raise ValueError("Semaphore initial value must be >= 0")
         self._value = value
-        self._num_ready = 0
         self._waiters = collections.deque()
         if loop is not None:
             self._loop = loop
@@ -417,7 +416,6 @@ class Semaphore(_ContextManagerMixin):
             waiter = self._waiters.popleft()
             if not waiter.done():
                 waiter.set_result(None)
-                self._num_ready += 1
                 return
 
     def locked(self):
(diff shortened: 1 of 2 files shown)


manipopopo:
Following the discussion of #269, the behavior of `asyncio.Semaphore` now obeys the semantic of the semaphore.
* There is no guarantee that *task A* will acquire the semaphore before *task B* if *A* called `semaphore.acquire()` before *B*.

Now the implementation of `Semaphore` is the same as `Queue`'s.

I am still thinking over `fut.cancel()` in `Semaphore.acquire`. Following the discussion of #269, I understand that the `Future` might be not done under some scenarios. According to the documentation of asyncio: (https://docs.python.org/3/library/asyncio-task.html)
>`result = await future` or `result = yield from future` – suspends the coroutine until the future is done ...

Will `assert fut.done()` be another choice which could let the user notice that there might be something wrong?

gvanrossum:
I need a little time to review this. I think you're on the right track though!

manipopopo:
I found that my comment above is wrong.
> the behavior of asyncio.Semaphore now ~~obeys~~ the semantic of the semaphore.

`asyncio.Semaphore` is not exactly the same as normal semaphore. The sequence of the tasks acquiring the `asyncio.Semaphore` is deterministic under some scenarios. 
For example, suppose there are only two tasks *A* and *B*. *A* is waiting for the semaphore `sem`. If we call `sem.acquire()` following `sem.release()` in task *B*, *B* will acquire `sem` before *A*. I'm not sure whether or not "changing the container of pending `Future`s to a list of randomly inserted elements" is superfluous.

Anyway, I should not say "`asyncio.Semaphore` now obeys the semantic of the semaphore."

asyncio/locks.py
```diff
@@ -425,18 +432,18 @@ def acquire(self):
...
-
-        fut = futures.Future(loop=self._loop)
-        self._waiters.append(fut)
-        try:
-            yield from fut
-            self._value -= 1
-            return True
-        finally:
-            self._waiters.remove(fut)
+        while self._value == 0:
```
gvanrossum: Can we make this <= 0? I know it could never be < 0, but somehow it still gives me the willies imagining that *if* somehow it could become < 0 and we'd never recover.

asyncio/locks.py
```diff
@@ -425,18 +432,18 @@ def acquire(self):
...
+            fut = futures.Future(loop=self._loop)
+            self._waiters.append(fut)
+            try:
+                yield from fut
+            except:
```
gvanrossum: I've changed my mind on this (again). The except clause should just catch CancelledException and the fut.cancel() is unnecesary.

asyncio/locks.py
manipopopo: Ok! I think we could 
* change the condition to `<= 0` and add comment above it. 
* leave the condition be == 0 and add a assertion in the while loop

Maybe the second choice can make programmers aware of the error. I'm not sure.

tests/test_locks.py
```diff
@@ -18,7 +17,6 @@
 
 
 class LockTests(test_utils.TestCase):
-
```
gvanrossum: Can you restore this blank line and all following?

tests/test_locks.py
```diff
@@ -783,22 +778,20 @@ def c4(result):
 
         test_utils.run_briefly(self.loop)
         self.assertEqual(0, sem._value)
-        self.assertEqual([1, 2, 3], result)
+        self.assertEqual(3, len(result))
```
gvanrossum: Can you change this to something like assertEqual({1, 2, 3}, set(result))?

asyncio/locks.py
gvanrossum: Also, can you add a comment saying something like "See similar code in Queue"?

gvanrossum:
Thanks! I've added some nits, let me know when you've addressed them and I'll merge this baby.

asyncio/locks.py
manipopopo: When I was testing `asyncio.Queue`, there was only one time that an exception other than `CancelledException` was catch: when a pending task waiting for a future is deleted, `GeneratorExit` is raised.

Do we need to add  `assert fut.done()` here?

gvanrossum:
Oh, one more thing. You need to have an account on bugs.python.org and you need to sign the online PSF contributor form: https://www.python.org/psf/contrib/contrib-form/ I will then check your status in the bug tracker so we know you've signed it. (This only needs to happen for your first contribution to Python. :-)

asyncio/locks.py
gvanrossum: No, I just want <= 0 and no comment. :-)

asyncio/locks.py
gvanrossum: Oh wow. Then please leave it as is. The GeneratorExit is worth a comment! This pretty much confirms the hunch I had about other exceptions possibly happening here...

tests/test_locks.py
manipopopo: But only `1` will definitely be in the list. The list may contains {1, 2, 3}, {1, 2, 4} or {1, 3, 4}.

manipopopo: @gvanrossum Do we need to add assertTrue(1 in result) here?

gvanrossum: Oh, I see. Then I think what you have is fine. (Boy this is a complex test. :-)

manipopopo:
@gvanrossum 
I changed the condition of the while loop in `acquire` and restored the blank lines. And I created an account on bugs.python.org and signed the form.

Thank you very much for the review!

manipopopo:
Oh and I think that when a pending task being deleted (raise `GeneratorExit` in `Semaphore.acquire`, `Queue.put` or `Queue.get`) there might be some bugs though ...

manipopopo:
The `generatorExit` will be raised when `close` of the genenator underlying the wrapping task is called. This may happen when the semaphore and the event loop (and all the other things reference to the task) are destroyed, and the *pending* task is going to be deleted. Maybe we don't need to worry about whether or not the state of the destroyed semaphore is correct.

On the other hand, the document says
> `result = await future` or `result = yield from future` – suspends the coroutine until the future is **done** ...

A pending task will be awoken by a callback function `_wakeup`, which will call `Task._step` of the task. And `_wakeup` will be scheduled if and only if the corresponding future is done. I am not sure whether or not "the spirit of `yield from future` implies that when we pass this line, we know the future is done".

gvanrossum:
We certainly can trust that when we pass `yield from future` *without raising an exception* the future is done. What we can't trust is whether it will be done when we catch an exeption, especially a BaseException instance (which is what GeneratorExit is). The philosophical question is then whether we might still care about the semaphore or queue when the generator is closed. I think it's at least conceivable that some code could create the generator, let it the event loop run for a little bit, and then decide to close the generator. So preserving the invariant of the queue or semaphore in this case sounds good. And that's what we do. So I'm going to merge your change!

gvanrossum:
Applied cc20f3951e81b8bbd405d9100c17a0b94c22b0b9.

I did a manual merge, so you may have to mark the PR as applied manually. Thanks!!

manipopopo:
I apologize for sticking to this issue for several days. I want to make sure whether or not *"the exceptions other than `CancelledException` should be take into account in other places using `await future` (or `yield from future`)."*

> ... some code could create the generator, let it the event loop run for a little bit, and then decide to close the generator ...

I think that a generator **G** (or a coroutine) calling `await future` must be wrapped in a Task **T** scheduled by the event loop. To close the generator, maybe the programmer should call `T.cancel()` instead of calling `close` of underlying **G** (if the programmer can reference **G**). I can't figure out under what scenario a user of *asyncio* is able to and would like to operate the wrapped **G** directly.

When I was testing `Queue`, I got `GeneratorExit` when the program was going to exit and there are still  some pending task which I didn't cancel before the program finished. I didn't know how to call `close` of the underlying generator without using `Task._coro`. As a user of *asyncio*, I think I should not use `_coro` with a leading underscore.
_____
>mark the PR as applied manually

Do I just leave a comment on this PR, or there are other steps I miss?

gvanrossum:
Here's a way that GeneratorExit can be caused:

```python
from asyncio import coroutine, sleep, ensure_future, get_event_loop

@coroutine
def foo():
    try:
        yield from sleep(1)
    except BaseException as err:
        print('hoho')

@coroutine
def bar():
    coro = foo()
    task = ensure_future(coro)
    yield from sleep(0.1)
    coro.close()

get_event_loop().run_until_complete(bar())
```
However I don't think that this is a very realistic scenario. Another scenario is when you receive a KeyboardInterrupt. Since that is a BaseException, it barrels right past `except Exception:` clauses elsewhere in asyncio.

Looking back at it I think that's probably a mistake -- when we use try/finally, the finally action definition happens even when a BaseException is raised, but more explicit cleanup in `except Exception:` blocks ignores it. It would be nice if all fundamental internal data structures remained in a consistent state even when BaseException (of whatever nature -- GeneratorExit, SystemExit or KeyboardInterrupt) left things in a consistent state, just in case some later recovery process fires up the event loop again.

PS. I think just closing the issue was enough to take the PR off the list.

manipopopo:
The thought that "calling `coro.close()` instead of `task.cancel()` in `bar` is a wrong use of _asyncio_" is what makes me stick to `GeneratorExit`. I think I'm too naive to imagine these things. Thank you very much for clarify all of these issues!!!

gvanrossum:
It *is* wrong to do that. But if we can write the code robustly, so that we don't care about wrong things people do, without making it unreadable or obfuscated or slower, we should write the code robustly. And I think that's what we've accomplished. Do you agree?

manipopopo:
Yes, I do. I learned a lot from the discussion! Thank you for all your help and guidance!

PR #273, opened 2015-09-28
Submitting a coroutine to an asyncio event loop

vxgmichel:
This PR follows up on this [conversation](https://mail.python.org/pipermail/python-ideas/2015-September/036512.html) on the python-ideas mailing list. It is based on the implementation I posted and commented [here](https://github.com/vxgmichel/asyncio-loopexecutor), but discarding the executor.

This PR includes:
- a `connect_futures` function, to safely connect two futures regardless of their types
- a `submit_to_loop` function, to submit a coroutine from a thread to a given event loop
- a test case for `submit_to_loop` (runs in about 15 ms)

It also affects `wrap_future`, refactored to use `connect_futures` and avoid code duplication.

I tried to gather my changes as much as possible. However, you might want to have a look at:
- `import asyncio` (in futures.py). I needed `ensure_future` for `submit_in_loop` and couldn't import it with `from . import tasks` (because of circular import).
- `_copy_state` (in futures.py). The code in `Future._copy_state` should probably move in `_copy_state` as I pointed out in a comment.
- `_schedule` works exactly like `ensure_future` but adds an optional `destination` argument. There could be some refactoring here, though you probably don't want to change the `ensure_future` prototype.

New names and prototypes might also need to be rethought, along with docstrings and type testing. I'm not sure either about how to use the `TestLoop`, though the tests seem to pass properly.


ajdavis:
Tornado's [chain_future](https://github.com/tornadoweb/tornado/blob/v4.2.1/tornado/concurrent.py#L475) is equivalent to your `connect_futures` so I propose using Tornado's name for this feature.

vxgmichel:
@ajdavis All right, let me know if you have other suggestions

gvanrossum:
Can you fill out the PSF contributor form at https://www.python.org/psf/contrib/contrib-form/ ? This is a one-time chore. :-)

asyncio/futures.py
```diff
@@ -387,25 +390,79 @@ def __iter__(self):
...
 
 
-def wrap_future(fut, *, loop=None):
+# Helpers to connect asyncio futures and concurrent futures
+
+def _copy_state(source, destination):
+    """Copy state from a future to another future.
+    Compatible with both asyncio.Future and concurrent.futures.Future."""
+    # Future._copy_state code should probably move in here
+    return Future._copy_state.__get__(destination)(source)
```
gvanrossum: Can you refactor this so that the actual code is here and Future._copy_state invokes this? Or even better make it just a helper function -- the one place where Future._copy_state is invoked you can just change it to call the function.

gvanrossum: (Actually I see you removed the direct call site. So you can just move the code here.)

asyncio/futures.py
```diff
@@ -387,25 +390,79 @@ def __iter__(self):
...
+    if isinstance(origin, concurrent.futures.Future) and \
+       isinstance(affected, Future):
+        affected._loop.call_soon_threadsafe(callback)
+    else:
+        callback()
+
+
+def chain_future(source, destination):
+    """Connect a future to another future.
+    Compatible with both asyncio.Future and concurrent.futures.Future."""
```
gvanrossum: Can you borrow a few words from the Tornado function's docstring to clarify what this does?

asyncio/futures.py
```diff
@@ -387,25 +390,79 @@ def __iter__(self):
...
+    if not isinstance(source, (Future, concurrent.futures.Future)):
+        raise TypeError('A future is required for source argument')
+    if not isinstance(destination, (Future, concurrent.futures.Future)):
+        raise TypeError('A future is required for destination argument')
+
+    def _check_cancel_other(destination):
+        if destination.cancelled():
+            _safe_callback(destination, source, source.cancel)
```
gvanrossum: I would prefer it if `_safe_callback()` didn't exist, and you just selected the right way to call the callback here based on the type of source and destination.

asyncio/futures.py
```diff
@@ -387,25 +390,79 @@ def __iter__(self):
...
+    Connect the result to an optional destination future."""
+    future = asyncio.ensure_future(coro, loop=loop)
+    if destination is not None:
+        chain_future(future, destination)
+    return future
+
+
+def submit_to_loop(coro, loop):
+    """Submit a coroutine to a given event loop.
+    Return a concurrent.futures.Future."""
```
gvanrossum: General nit: please follow the docstring style for multi-line docstrings from the rest of the package; a blank line after the initial line, and the final """ on a line by itself.

asyncio/futures.py
```diff
@@ -387,25 +390,79 @@ def __iter__(self):
...
-    def _check_cancel_other(f):
-        if f.cancelled():
-            fut.cancel()
 
-    new_future.add_done_callback(_check_cancel_other)
-    fut.add_done_callback(
-        lambda future: loop.call_soon_threadsafe(
-            new_future._copy_state, future))
-    return new_future
+# Submission of a coroutine to an event loop
```
gvanrossum: Because this code depends on `ensure_future()` it should live in tasks.py.

asyncio/futures.py
```diff
@@ -10,9 +11,11 @@
 import reprlib
 import sys
 import traceback
+import functools
 
 from . import compat
 from . import events
+import asyncio
```
gvanrossum: I really don't like this import. The code that necessitates it probably should be moved to tasks.py.

asyncio/futures.py
```diff
@@ -387,25 +390,79 @@ def __iter__(self):
...
+    return future
+
+
+def submit_to_loop(coro, loop):
+    """Submit a coroutine to a given event loop.
+    Return a concurrent.futures.Future."""
+    if not asyncio.iscoroutine(coro):
+        raise TypeError('A coroutine is required')
+    future = concurrent.futures.Future()
+    callback = functools.partial(_schedule, coro, loop=loop,
```
gvanrossum: Nit: I prefer a lambda instead of partial(). (Possibly `_schedule()` can be made a local helper function here?

gvanrossum:
Thanks! This looks promising. I've left a bunch of refactoring suggestions. Let me know when you're ready for another code review.

asyncio/futures.py
vxgmichel: Thanks for the review!
I just pushed a [commit](https://github.com/vxgmichel/asyncio/commit/7bafb683bc184eb9dbe8fdf4180c3d39929042cb) to make `_copy_state` a static method, is it ok? I had to update the tests and I noticed [this line](https://github.com/python/asyncio/blob/master/tests/test_futures.py#L177):
```
# Test the internal _copy_state method since it's being directly
# invoked in other modules.
````
I couldn't find any other usage in `asyncio`, is another project using this method?

Comment (gvanrossum) [inline asyncio/futures.py]: That comment may be old, but it's best to have an explicit test for this function anyway.

Anyway, I hate static methods, can you just make it a function?

**Comment by vxgmichel** [comment]
All right, it should look better now!
I also changed the way `chain_future` decides whether to use `call_soon_threadsafe` or not. Now it compares the loops instead of checking the type of the futures.

File: asyncio/futures.py [hunk 8]
```diff
@@ -360,25 +360,6 @@ def set_exception(self, exception):
 
     # Truly internal methods.
```
gvanrossum: This comment is obsolete; just delete it.

asyncio/futures.py
```diff
@@ -387,25 +368,66 @@ def __iter__(self):
...
-    new_future = Future(loop=loop)
+def _copy_state(source, destination):
+    """Copy state from a future to another future.
+
+    Compatible with both asyncio.Future and concurrent.futures.Future.
+    """
+    assert source.done()
+    if destination.cancelled():
+        return
+    assert not destination.done()
```
gvanrossum: Something just occurred to me. If the destination is a concurrent.futures.Future, don't we have a race condition here? The destination may not be cancelled on the previous line, but it might still be cancelled (and hence done()) by the time we reach this line. Now, concurrent.futures.Future doesn't mind if you set the result multiple times or after it's been cancelled. So perhaps we should just drop the assert? (The cancelled() check is still needed, in case the destination is an asyncio.Future.)

asyncio/futures.py
```diff
@@ -387,25 +368,66 @@ def __iter__(self):
...
+    if not isinstance(source, (Future, concurrent.futures.Future)):
+        raise TypeError('A future is required for source argument')
+    if not isinstance(destination, (Future, concurrent.futures.Future)):
+        raise TypeError('A future is required for destination argument')
+    source_loop = source._loop if isinstance(source, Future) else None
+    dest_loop = destination._loop if isinstance(destination, Future) else None
+
+    def _check_cancel_other(destination):
+        if destination.cancelled():
+            if source_loop is not None and source_loop != dest_loop:
```
gvanrossum: Instead of != use `is not` (they mean the same thing but the intention is to compare the identities). Same below.

asyncio/tasks.py
```diff
@@ -680,3 +680,19 @@ def _done_callback(inner):
...
     return outer
+
+
+def submit_to_loop(coro, loop):
+    """Submit a coroutine to a given event loop.
+
+    Return a concurrent.futures.Future to access the result.
+    """
+    if not coroutines.iscoroutine(coro):
+        raise TypeError('A coroutine is required')
```
gvanrossum: Can you make that "A coroutine object is required" ? ("Coroutine" could refer to either a coroutine function or a coroutine object -- but we only allow the latter.)

gvanrossum:
I like the new refactoring a lot! Just a few nits left.

1st1:
Couple of comments:

1. `submit_to_loop` is a very generic name, but the current implementation accepts only coroutines and returns an instance of `concurrent.Future`.  Can we have a better, more descriptive name?

2. Do we have to expose the `chain_future` function at all?

asyncio/futures.py
```diff
@@ -387,25 +368,66 @@ def __iter__(self):
         return self.result()  # May raise too.
 
     if compat.PY35:
-        __await__ = __iter__ # make compatible with 'await' expression
+        __await__ = __iter__  # make compatible with 'await' expression
```
1st1: We don't need this change as part of this commit...

asyncio/futures.py
vxgmichel: Oops, there's indeed a concurrency problem with this piece of code. The `assert` should definitely be removed, but I'm also a bit surprised that going from `CANCELLED` to `FINISHED` is a legal transition for a `concurrent.futures.Futures`. If this is the intended behavior, maybe we should change the code to:
```python
if isinstance(destination, Future) and destination.cancelled():
        return
```
so that the result is always set on the cancelled `concurrent.futures.Future` if the source manages to terminate its task before `source.cancel` is called. What do you think?

gvanrossum: The secret handshake is set_running_or_notify_cancel(). If this returns True you can set the result/exception; otherwise don't bother.

I wonder if we just need to have a different _copy_state() implementation if the destination is a concurrent.futures.Future. All this "works with either type of future" stuff feels scary and wrong and not a pattern we'd want to encourage in users -- they should keep their Futures separated and not confuse them (despite the similar interfaces).

vxgmichel:
@1st1 
 1. Sure, what about `submit_coroutine`? It implies that the coroutine will run in an event loop (since it is the only place it can run in) and `submit` refers to `Executor.submit` that returns a `concurrent.futures.Future`.

 2. `chain_future` is a generic version of `wrap_future`, and since `chain_future` already exists in [tornado](https://github.com/tornadoweb/tornado/blob/v4.2.1/tornado/concurrent.py#L475), I thought it made sense to expose it. However, it is true that `chain_future` and `wrap_future` are not directly useful since `submit_coroutine` and `run_in_executor` should cover most of the use cases.

asyncio/futures.py
vxgmichel: This implementation prevents the result to be potentially set on a cancelled `concurrent.futures.Future`:

```python
def _copy_state(source, destination):
    assert source.done()
    if isinstance(destination, concurrent.futures.Future):
        destination.set_running_or_notify_cancel()
    if destination.cancelled():
        return
    if source.cancelled():
        destination.cancel()
    else:
        [...]  # Copy result or exception
```
If you want to separate the implementations, it would be nice to have `concurrent.futures.Future` implement its own `_copy_state` method so that chain future can reliably use `destination._copy_state(source)`.

1st1:
@vxgmichel @gvanrossum 

> Sure, what about submit_coroutine? It implies that the coroutine will run in an event loop (since it is the only place it can run in) and submit refers to Executor.submit that returns a concurrent.futures.Future.

I like `submit_coroutine` better, but still, it doesn't reflect that an instance of a `concurrent.futures.Future` is returned.  If I'm not mistaken we don't have any other API that returns non-asyncio futures.

How about `schedule_concurrent` or `schedule_coroutine_concurrently` or `run_coroutine_concurrently`?  They are very verbose, but descriptive.

> chain_future is a generic version of wrap_future, and since chain_future already exists in tornado, I thought it made sense to expose it. However, it is true that chain_future and wrap_future are not directly useful since submit_coroutine and run_in_executor should cover most of the use cases.

It was requested by someone on bugs.python.org to better document `wrap_future`, since we decided not to support awaiting on concurrent futures in asyncio. I think it's a valid use case, in fact there is an issue to document this function.

`chain_future` is hardcoded to only accept concurrent futures and asyncio futures, it won't accept one from Tornado, for instance.  It's implementation is rather complex, and it's hard to understand what it does.  I'd really prefer to keep this function private if possible.

asyncio/futures.py
vxgmichel: Sorry, that was my PEP 8 twitch :smile:

asyncio/futures.py
gvanrossum: Let's not add new private methods to the concurrent package. Also, you need to check the result of destination.set_running_or_notify_cancel().

vxgmichel: Well, I thought `if destination.cancelled(): return` achieves the same thing. Would you still prefer:
```python
    if isinstance(destination, concurrent.futures.Future):
        if not destination.set_running_or_notify_cancel():
            return
    elif destination.cancelled():
        return
```

gvanrossum: Yes.

On Wed, Sep 30, 2015 at 11:07 AM, Vincent Michel <notifications@github.com>
wrote:

> In asyncio/futures.py
> <https://github.com/python/asyncio/pull/273#discussion_r40828964>:
>
> > -    if isinstance(fut, Future):
> > -        return fut
> > -    assert isinstance(fut, concurrent.futures.Future), \
> > -        'concurrent.futures.Future is expected, got {!r}'.format(fut)
> > -    if loop is None:
> > -        loop = events.get_event_loop()
> > -    new_future = Future(loop=loop)
> > +def _copy_state(source, destination):
> > +    """Copy state from a future to another future.
> > +
> > +    Compatible with both asyncio.Future and concurrent.futures.Future.
> > +    """
> > +    assert source.done()
> > +    if destination.cancelled():
> > +        return
> > +    assert not destination.done()
>
> Well, I thought if destination.cancelled(): return achieves the same
> thing. Would you still prefer:
>
>     if isinstance(destination, concurrent.futures.Future):
>         if not destination.set_running_or_notify_cancel():
>             return
>     elif destination.cancelled():
>         return
>
> —
> <https://github.com/python/asyncio/pull/273/files#r40828964>.
>



-- 

vxgmichel:
@1st1 @gvanrossum 
> How about schedule_concurrent or schedule_coroutine_concurrently or run_coroutine_concurrently? They are very verbose, but descriptive.

I'll let you guys settle on a name, I changed it to `submit_coroutine` for the moment.

> I'd really prefer to keep this function private if possible.

All right, I removed `chain_future` from `__all__`. Should we rename it to `_chain_future` as well?

asyncio/futures.py
```diff
@@ -390,22 +369,66 @@ def __iter__(self):
...
+    """
+    assert source.done()
+    if isinstance(destination, concurrent.futures.Future):
+        if not destination.set_running_or_notify_cancel():
+            return
+    elif destination.cancelled():
+        return
+    assert not destination.done()
+    if source.cancelled():
+        destination.cancel()
```
gvanrossum: This actually doesn't work for a concurrent Future. Its cancel() method is a no-op once the state is RUNNING (which it is once set_running_or_notify_cancel() returns).

asyncio/futures.py
```diff
@@ -390,22 +369,66 @@ def __iter__(self):
...
+    else:
+        exception = source.exception()
+        if exception is not None:
+            destination.set_exception(exception)
+        else:
+            result = source.result()
+            destination.set_result(result)
+
+
+def chain_future(source, destination):
```
gvanrossum: Now that we're not exporting this, please add a _ to the front of the name.

asyncio/tasks.py
```diff
@@ -680,3 +680,19 @@ def _done_callback(inner):
 
     inner.add_done_callback(_done_callback)
     return outer
+
+
+def submit_coroutine(coro, loop):
```
gvanrossum: Executive decision: Let's call this `run_coroutine_threadsafe()`. The `_threadsafe()` suffix resembles that of `call_soon_threadsafe()` (which is the only other thing that's safe to call from a different thread).

1st1: Should we consider adding a keyword-only argument to specify if it should return an asyncio or concurrent future?

gvanrossum: Please no. There is already way too much dynamic type checking in this code than I'm comfortable with -- I want less of that, not more. The use case for submitting from one loop to another is entirely hypothetical at this point -- but the use case for submitting from an executor to a loop was clearly documented by Vincent in the thread that led to this PR.

asyncio/futures.py
vxgmichel: You're right, this implementation of `_copy_state`  is trickier than what I expected... I  came up with this:
```python
def _copy_state(source, destination):
    """Copy state from a future to another future.

    Compatible with both asyncio.Future and concurrent.futures.Future.
    """
    # Check source cancellation
    assert source.done()
    if source.cancelled():
        destination.cancel()
        return
    # Check destination cancellation
    if isinstance(destination, concurrent.futures.Future):
        if not destination.set_running_or_notify_cancel():
            return
    elif destination.cancelled():
        return
    assert not destination.done()
    # Set exception or result
    exception = source.exception()
    if exception is not None:
        destination.set_exception(exception)
    else:
        result = source.result()
        destination.set_result(result)
```

It should be ok to check the source cancellation before since a `cancel` method should never fail, regardless of the future type. I also wrote a new test to cover the case you described.

asyncio/tasks.py
vxgmichel: Just a random thought, is there a reason for not making it a `BaseEventLoop` method?
```python
loop.run_coroutine_threadsafe(coro) # similar to
loop.call_soon_threadsafe(callback)
```

1st1: I'd keep it separate since it's not something that will be used frequently and because it returns a non-asyncio future object.

gvanrossum: This had occurred to me too, but looking into the reason *why* `call_soon_threadsafe()` is a loop method, I see that it is because it needs to use a mechanism internal to each loop implementation in order to wake up the loop. (Note that there are multiple implementations of `_write_to_self()`.)

That reasoning doesn't apply here. So I'd rather have this as a helper function.

asyncio/futures.py
vxgmichel: @gvanrossum 
I thought about what you said and separating the two `_copy_state` implementations. What about keeping the current implementation of `Future._copy_state` and changing `_copy_state` to:
```python
def _copy_state(source, destination):
    # Use Future._copy_state implementation
    if isinstance(destination, Future):
        return destination._copy_state(source)
    # Implements _copy_state for concurrent.future.Futures
    assert source.done()
    if source.cancelled():
        destination.cancel()
    if not destination.set_running_or_notify_cancel():
        return
    exception = source.exception()
    if exception is not None:
        destination.set_exception(exception)
    else:
        result = source.result()
        destination.set_result(result)
```
Pros:
- Each implementation can be maintained separately 
- The second part can easily be moved somewhere else if needed

Cons:
- A bit of code duplication

gvanrossum: I like this idea. I would go even further -- I would have a `_copy_asyncio_future_state_to_concurrent_future()` method (maybe use a shorter name :-) and decide which one to use when registering the callback.

vxgmichel: Actually, the source type doesn't matter since the source is done already: the only thing we need to know about it is if it's cancelled, done with a result or done with an exception. This is why `_copy_state` fits very well as a method, since the only type that matters is the type of the instance itself. I thought of `_set_concurrent_future_state` with the following prototype:
```python
def _set_concurrent_future_state(concurrent, other):
```
I think it's less confusing and more similar to:
```python
    def _copy_state(self, other):
```
I'll push it in a second, please let me know what you think about it.

gvanrossum: Sounds good, but please name the other argument `source`.

vxgmichel: All right, should I rename it in `Future._copy_state` too for consistency?

gvanrossum: No, please leave that alone. Consistency is overrated. ;-)

gvanrossum:
Thanks! I consider this done except for Python docs. I've manually applied your patch and pushed it to the repo. I'm in the middle of committing the patch to the CPython repo. I would love it if you volunteered doc updates for CPython! (Starting with 3.4.) See you in bugs.python.org for that...

gvanrossum:
Cross-reference CPython issue: http://bugs.python.org/issue25304

PR #274, opened 2015-09-28
Make peer PID available via transport.get_extra_info for UNIX domain sockets

neumond:
It wasn't obvious how to get connected peer's PID. get_extra_info('peername') returns empty bytes object.

### DIFF
--- a/asyncio/selector_events.py
+++ b/asyncio/selector_events.py
@@ -514,6 +514,8 @@ class _SelectorTransport(transports._FlowControlMixin,
                 self._extra['peername'] = sock.getpeername()
             except socket.error:
                 self._extra['peername'] = None
+        if sock.family == socket.AF_UNIX:
+            self._extra['peercred'] = sock.getsockopt(socket.SOL_SOCKET, socket.SO_PEERCRED)
         self._sock = sock
         self._sock_fd = sock.fileno()
         self._protocol = protocol


haypo:
It's easy to get the Python socket object from the transport: sock = transport.get_extra_info('socket').
https://docs.python.org/dev/library/asyncio-protocol.html#asyncio.BaseTransport.get_extra_info

Why not getting the peer credential from this object?

```
sock = transport.get_extra_info('socket')
peercred = sock.getsockopt(socket.SOL_SOCKET, socket.SO_PEERCRED)
```

gvanrossum:
I agree.

neumond:
Isn't that a purpose of get_extra_info to keep most requested data for transport?

PID of process is analogous to the 'peername' address in case of network connections. Unix server can accept many connections and I need to distinguish them, especially if I make a master process spawning multiple children.

Also, of course, you see it very easy since you just had a look at my code. But I couldn't figure out that without 20min googling and experimenting. And many other non socket pro people too.

haypo:
Your code looks wrong, the recipes found on the Internet pass a length to getsockopt() and then unpack a structure:

```
creds = conn.getsockopt(SOL_SOCKET, SO_PEERCRED, struct.calcsize('3i'))
pid, uid, gid = struct.unpack('3i', creds)
```

A raw bytes string is not the best API for asyncio.

I'm not sure that the structure is portable neither.

haypo:
> I'm not sure that the structure is portable neither.

Oh, it looks like SO_PEERCRED is specific to Linux. On FreeBSD, it's only emulated for compatibility with Linux applications. It looks like it's not available on Solaris, MacOS X, nor OpenBSD.

neumond:
Well. My code works hacky way, returning one int (bufsize=0) used to take socket flags, not arrays of int.

Implementing portable way to get peercred will probably require changes in Python source (I think its better place to more people to benefit from it).

I guess it is worth to look in postgres source since it's access rules work based on peercred. And this gives assurance that someone even uses heavilty peercred functionality.

MacOS:
https://developer.apple.com/library/prerelease/mac/documentation/Darwin/Reference/ManPages/man3/getpeereid.3.html
For FreeBSD, looks like it called LOCAL_PEERCRED. I didn't get what you meant by "emulated".

Anyway, I think it's better to look in postgres code (downloading it now) and take ready code.

Issue #277, opened 2015-10-01
Support both async context manager and awaitable in the same place


asvetlov:
This is long enough story, please be patient.

I've added pep492 support to aiohttp client api.

Prelude
----------

Initially we had client request call like

    client = aiohttp.ClientSession(loop=loop)
    resp = yield from client.request('GET', url)

or even 

    resp = yield from aiohttp.get(url)

We support bare `yield from aiohttp.request('GET', url)` and family also but these shortcuts work over `ClientSession` now so let's forget about shortcuts.

I've added next step:

    resp = await client.get(url)
    async with resp:
        ...  # do your work on response object

The motivation is: HTTP response should be closed gracefully (`await resp.release()`) to support HTTP keep-alives. Everything works fine.

But I would have even more convenient API:

    async with client.get(url) as resp:
         ...

with keeping ability to use old-good `resp = await client.get(url)`. 
Backward compatibility matters and so on.

Consequently `.get(...)` method cannot be coroutine but should be regular function returning an object which is *support both protocols: coroutine and async context manager*.

Following the way I've got something like:

    def get(self, url):
         return _ContextManager(self._request('GET', url))

where `self._request(...)` is actual coroutine which performs HTTP request and

    class _ContextManager:
         def __init__(self, coro):
               self._coro = coro

         @asyncio.coroutine
         def __iter__(self):
               yield from self._coro

         def __await__(self):
               yield from self._coro

         @asyncio.coroutine
         def __aenter__(self):
               self._resp = yield from self._coro

         @asyncio.coroutie
         def __aexit__(self, exc, val, tb):
              if exc is None:
                  yield from self._resp.release()
              else:
                   # on exception we can just drop connection without returning response to pool
                   self._resp.close()  

Thank you for reading up to the point.

Problem
------------ 

in `aiohttp` docs we declared the `loop.run_until_complete(client.get(url))` is a legal construction.

`run_until_complete` calls `ensure_future()` which do perform checks for `isinstance(_ContextManager, Future)` (obviously not) and `iscoroutine(_ContextManager(...))`.

Well, I've found I should derive `_ContextManager` from `collections.abc.Coroutine` and implement  `.send`, `.throw` and `.close`. 

There is my final `_ContextManager` implementation https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/client.py#L422-L464 

[...]

PR #278, opened 2015-10-02
Make ensure_future() accept awaitables in 3.5

1st1:
See also issue #277.

Since `ensure_future()` is used by `loop.run_until_complete()`, I think it's a good idea to enable it for all awaitable objects (the ones that implement `__await__` in Python 3.5 and later).  With this PR it's valid to write `loop.run_until_complete(awaitable)`.

### DIFF
--- a/asyncio/tasks.py
+++ b/asyncio/tasks.py
@@ -512,7 +512,7 @@ def async(coro_or_future, *, loop=None):
 
 
 def ensure_future(coro_or_future, *, loop=None):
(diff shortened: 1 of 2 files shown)


gvanrossum:
Looks good!

1st1:
@gvanrossum Thanks!

asvetlov:
I believe documentation for `ensure_future()` should be updated accordingly.

Issue #282, opened 2015-11-02
ABC for Futures


1st1:
I'm experimenting with implementing an asyncio-compatible event loop on top of libuv with Cython. It would be great if there is a way to provide a custom Future implementation.  I was thinking about adding `loop.create_future()` (instead of `Future(loop=loop)`) API method, as well as an ABC, so that functions like `ensure_future` would recognize custom Futures.  @gvanrossum @haypo what do you think about this?

gvanrossum:
Hm... Why can't your new loop reuse the existing Future class? Or subclass
it?

On Sun, Nov 1, 2015 at 4:45 PM, Yury Selivanov <notifications@github.com>
wrote:

> I'm experimenting with implementing an asyncio-compatible event loop on
> top of libuv with Cython. It would be great if there is a way to provide a
> custom Future implementation. I was thinking about adding
> loop.create_future() (instead of Future(loop=loop)) API method, as well
> as an ABC, so that functions like ensure_future would recognize custom
> Futures. @gvanrossum <https://github.com/gvanrossum> @haypo
> <https://github.com/haypo> what do you think about this?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/python/asyncio/issues/282>.
>



-- 
--Guido van Rossum (python.org/~guido)

1st1:
I can reuse the existing Future (but I can't derive extension types from Python classes in Cython). I just wanted to make a little bit faster `Future`, and, especially, `Task` classes.

Maybe I'm trying to micro-optimize here.

haypo:
Hack asyncio in a local branch, come with a microbenchmark and then we can
discuss ;-)

Once I implemented a free list for Task. The gain was null, so I gave up.

1st1:
@haypo Yeah, I'm closing this issue for now.  I'll come back with a microbenchmark if there is any difference.

1st1:
@gvanrossum @haypo @asvetlov So now I have benchmarks ;)

Long story short, here is an implementation of asyncio-compatible event loop on top of libuv: https://github.com/1st1/uvloop.  It's an early preview, however, at this point enough is implemented to run various benchmarks.  I think that I got most of the low-level implementation right, now I need to add more tests and implement the missing APIs.

For benchmarks I used [the ones](https://github.com/dabeaz/curio/tree/master/examples/bench) that David Beazley used for his curio project.  

#### Echo server using low-level `sock_recv` / `sock_sendall`

[...]

PR #289, opened 2015-11-13
Optimize Task._step and Task._wakeup

1st1:
I found a way to [monkey-patch](https://github.com/1st1/uvloop/commit/116830fa9db02960aaf78066afdb5d6d76d33a00#diff-4bfb418c867e6a8b43257755c7bdaaccR13) `asyncio.Future` and make it compatible with uvloop (see #282 for details).

This PR will significantly improve futures performance in uvloop (**~10-15%**) while we wait till 3.5.2.  In 3.5.2, instead of adding `AbstractFuture`, or using `ABCMeta`, I propose to directly incorporate my C accelerator in asyncio -- that's probably the simplest solution.

This change is 100% backwards compatible -- it's a pure optimization.  Please consider accepting this before 3.5.1.

### DIFF
--- a/asyncio/tasks.py
+++ b/asyncio/tasks.py
@@ -247,7 +247,8 @@ class Task(futures.Future):
             self.set_exception(exc)
             raise
         else:
-            if isinstance(result, futures.Future):
+            # Use duck-typing here. `isinstance(result, Future)` is slow.
+            if hasattr(result, 'add_done_callback'):
                 # Yielded Future must come from Future.__iter__().
                 if result._blocking:
                     result._blocking = False
@@ -290,7 +291,15 @@ class Task(futures.Future):
             # This may also be a cancellation.
             self._step(None, exc)
         else:
-            self._step(value, None)
+            # We don't need to pass the `value` to `_step` here.
+            # When `_step` is called, it will call `coro.send(None)`,
+            # which will, in turn, call Future's iterator (returned from
+            # Future.__iter__ or Future.__await__) __next__.  If you push
+            # a non-None value here, 'send' method of that iterator
+            # would be called (instead of __next__) -- this is slow for
+            # Future implementations that do not return a generator from
+            # their __iter__.
+            self._step()
         self = None  # Needed to break cycles when an exception occurs.
 
 


asyncio/tasks.py
```diff
@@ -247,7 +247,8 @@ def _step(self, value=None, exc=None):
             self.set_exception(exc)
             raise
         else:
-            if isinstance(result, futures.Future):
+            # Use duck-typing here. `isinstance(result, Future)` is slow.
+            if hasattr(result, 'add_done_callback'):
```
gvanrossum: I would recommend checking for a method that concurrent.futures.Future does not have.

asyncio/tasks.py
```diff
@@ -290,7 +291,15 @@ def _wakeup(self, future):
...
         else:
-            self._step(value, None)
+            # We don't need to pass the `value` to `_step` here.
+            # When `_step` is called, it will call `coro.send(None)`,
+            # which will, in turn, call Future's iterator (returned from
+            # Future.__iter__ or Future.__await__) __next__.  If you push
+            # a non-None value here, 'send' method of that iterator
+            # would be called (instead of __next__) -- this is slow for
+            # Future implementations that do not return a generator from
+            # their __iter__.
```
gvanrossum: This confused me. Are you saying that the `coro.send(value)` would end up resuming the `yield self` in `Future.__iter__()`? In that case I agree that the value is unnecessary since it is ignored there. And I suppose the cost is that looking up and calling some alternative Future's `send()` is more expensive than using the builtin `__iter__()`? Once thing seems to be missing though -- the code at line 239 is still calling `coro.send(None)` -- where does the translation to `__iter__()` happen?

1st1: @gvanrossum 

> Are you saying that the coro.send(value) would end up resuming the yield self in Future.__iter__()? In that case I agree that the value is unnecessary since it is ignored there.

Yes, exactly.  `Future.__iter__` will `return` the result itself.

> And I suppose the cost is that looking up and calling some alternative Future's send() is more expensive than using the builtin __iter__()?

Yes, instead of `__next__`, ceval.c will call `send`.  Since __next__ can be implemented as a slot, it's faster.

> [..]where does the translation to __iter__() happen?

ceval will call `__next__()` if `coro.send(None)`, and `send(val)` if `coro.send(val)`

asyncio/tasks.py
1st1: On a macro benchmark the difference is about 4%. Let me come up with a micro benchmark...

asyncio/tasks.py
1st1: > ceval will call __next__() if coro.send(None), and send(val) if coro.send(val)

This, btw, would be nice to fix in Python 3.6 -- we can add a slot for `send` in `tp_as_async` (or a new slot tp_as_generator)

gvanrossum: I presume the code that chooses between `send()` or `__next__()` is in
`gen_send_ex()` in `genobject.c`? Honestly I don't see where it's doing
this. `Task._send()` always calls `coro.send(value)` -- maybe you meant to
change it here???

Re: making `send()` into a slot: that seems highly unusual, since the way
it's invoked is through a regular method call, not through a special
operator (syntax).

1st1: > I presume the code that chooses between `send()` or `__next__()` is in
`gen_send_ex()` in `genobject.c`? Honestly I don't see where it's doing
this.

It's happening in `ceval.c`:

```c
        TARGET(YIELD_FROM) {
            PyObject *v = POP();
            PyObject *reciever = TOP();
            int err;
            if (PyGen_CheckExact(reciever) || PyCoro_CheckExact(reciever)) {
                retval = _PyGen_Send((PyGenObject *)reciever, v);
            } else {
                _Py_IDENTIFIER(send);
                if (v == Py_None)
                    retval = Py_TYPE(reciever)->tp_iternext(reciever);
                else
                    retval = _PyObject_CallMethodIdObjArgs(reciever, &PyId_send, v, NULL);
            }
```

1st1: > Re: making `send()` into a slot: that seems highly unusual, since the way
it's invoked is through a regular method call, not through a special
operator (syntax).

I'd say it's highly unusual for `ceval.c` to invoke a `__dict__` method instead of a slot.

asyncio/tasks.py
1st1: On a micto benchmark the difference is about 4% too.

After giving this some thought, I think we can keep this isinstance, since, quoting you, "[...] checking for a method that concurrent.futures.Future does not have" I don't like -- feels too hacky.

Can I merge the second change `_step(value, None) -> _step()`, though?

gvanrossum: You need to slow down.

1st1:
@gvanrossum So what do you think about this PR?

gvanrossum:
OK, you can commit the change to the `_step()` call.

Issue #287, opened 2015-11-12
Wrong exception re-raises if task raises CancelledError inside except block


germn:
```python
import asyncio
from contextlib import suppress


async def main():
    task = asyncio.ensure_future(asyncio.sleep(1))
    task.cancel()

    try:
        raise Exception()
    except Exception:
        with suppress(asyncio.CancelledError):
            await task
        raise

if __name__ == "__main__":
    loop = asyncio.get_event_loop()
    loop.run_until_complete(main())
```
Expected output:
```python
Exception
```
Actual output:
```python
concurrent.futures._base.CancelledError
```
Windows 10, Python 3.5.0

gvanrossum:
I can repro this too -- it's not unique to Windows or to PEP 492 (it fails
the same way with coroutine/yield-from in Python 3.4). It's also not due to
something in contextlib.suppress() or due to having a try/except in the
outer except clause.

@1st1 -- do you have any suggestions? I suspect there's some special-casing
of CancelledError that goes wrong here.

1st1:
I might be mistaken here, but it looks like that it's a bug in CPython...

```python
class MainError(Exception):
    pass

class SubError(Exception):
    pass

def sub():
    yield

def main():
    try:
        raise MainError()
    except MainError:
        try:
            yield from sub()
        except SubError:
            pass
        raise

coro = main()
coro.send(None)
coro.throw(SubError())
```

Outputs:

```
Traceback (most recent call last):
  File "t.py", line 47, in <module>
    coro.throw(SubError())
  File "t.py", line 43, in main
    raise
RuntimeError: No active exception to reraise
```

1st1:
And if you replace `except MainError:` with `except MainError as e:`, and `raise` with `raise e` then everything works correctly.

1st1:
It's not even a `yield from` bug:

```python
class MainError(Exception):
    pass

class SubError(Exception):
    pass

def main():
    try:
        raise MainError()
    except MainError:
        try:
            yield
        except SubError:
            print('got SubError')
        raise

coro = main()
coro.send(None)
coro.throw(SubError())
```

Outputs:

```
got SubError
Traceback (most recent call last):
  File "t.py", line 19, in <module>
    coro.throw(SubError())
  File "t.py", line 15, in main
    raise
RuntimeError: No active exception to reraise
```

1st1:
http://bugs.python.org/issue25612

gvanrossum:
Thanks for the thorough investigation!

PR #292, opened 2015-11-16
Cleanup Future API

1st1:
This PR cleans up Future class by factoring out `_copy_state()` and `_set_result_unless_cancelled()` into separate functions.

### DIFF
--- a/asyncio/futures.py
+++ b/asyncio/futures.py
@@ -20,7 +20,7 @@ _CANCELLED = 'CANCELLED'
 _FINISHED = 'FINISHED'
 
 Error = concurrent.futures._base.Error
-CancelledError = concurrent.futures._base.CancelledError
+CancelledError = concurrent.futures.CancelledError
 TimeoutError = concurrent.futures.TimeoutError
 
 STACK_DEBUG = logging.DEBUG - 1  # heavy-duty debugging
@@ -154,7 +154,7 @@ class Future:
         if self._loop.get_debug():
             self._source_traceback = traceback.extract_stack(sys._getframe(1))
 
-    def _format_callbacks(self):
+    def __format_callbacks(self):
         # Private method, do not rely on its existence.
         cb = self._callbacks
         size = len(cb)
@@ -186,7 +186,7 @@ class Future:
                 result = reprlib.repr(self._result)
                 info.append('result={}'.format(result))
         if self._callbacks:
-            info.append(self._format_callbacks())
+            info.append(self.__format_callbacks())
         if self._source_traceback:
             frame = self._source_traceback[-1]
             info.append('created at %s:%s' % (frame[0], frame[1]))
@@ -355,8 +355,6 @@ class Future:
             # have had a chance to call result() or exception().
             self._loop.call_soon(self._tb_logger.activate)
 
-    # Truly internal methods.
-
     def __iter__(self):
         if not self.done():
             self._blocking = True
@@ -390,7 +388,7 @@ def _set_concurrent_future_state(concurrent, source):
         concurrent.set_result(result)
 
 
-def _apply_future_state(source, dest):
+def _copy_future_state(source, dest):
     """Internal helper to copy state from another Future.
 
     The other Future may be a concurrent.futures.Future.
@@ -426,7 +424,7 @@ def _chain_future(source, destination):
 
     def _set_state(future, other):
         if isinstance(future, Future):
-            _apply_future_state(other, future)
+            _copy_future_state(other, future)
         else:
             _set_concurrent_future_state(future, other)
 
--- a/asyncio/futures.py
+++ b/asyncio/futures.py
@@ -175,6 +175,7 @@ class Future:
         return 'cb=[%s]' % cb
 
     def _repr_info(self):
+        # Private method, do not rely on its existence.
         info = [self._state.lower()]
         if self._state == _FINISHED:
             if self._exception is not None:
@@ -321,12 +322,6 @@ class Future:
 
     # So-called internal methods (note: no set_running_or_notify_cancel()).
 
-    def _set_result_unless_cancelled(self, result):
-        """Helper setting the result only if the future was not cancelled."""
-        if self.cancelled():
-            return
-        self.set_result(result)
-
     def set_result(self, result):
         """Mark the future done and set its result.
 
@@ -371,6 +366,13 @@ class Future:
 
     if compat.PY35:
         __await__ = __iter__ # make compatible with 'await' expression
+
+
+def _set_result_unless_cancelled(fut, result):
+    """Helper setting the result only if the future was not cancelled."""
+    if fut.cancelled():
+        return
+    fut.set_result(result)
 
 
 def _set_concurrent_future_state(concurrent, source):
--- a/asyncio/futures.py
+++ b/asyncio/futures.py
@@ -155,7 +155,7 @@ class Future:
             self._source_traceback = traceback.extract_stack(sys._getframe(1))
 
     def __format_callbacks(self):
-        # Private method, do not rely on its existence.
+        # Private to the asyncio module; unsafe to use in subclasses.
         cb = self._callbacks
         size = len(cb)
         if not size:
@@ -175,7 +175,7 @@ class Future:
         return 'cb=[%s]' % cb
 
     def _repr_info(self):
-        # Private method, do not rely on its existence.
+        # Private to the asyncio module; unsafe to use in subclasses.
         info = [self._state.lower()]
         if self._state == _FINISHED:
             if self._exception is not None:
@@ -230,7 +230,7 @@ class Future:
         return True
 
     def _schedule_callbacks(self):
-        # Private method, do not rely on its existence.
+        # Private to the asyncio module; unsafe to use in subclasses.
         """Internal: Ask the event loop to call all callbacks.
 
         The callbacks are scheduled to be called as soon as possible. Also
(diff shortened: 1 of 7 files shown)


asyncio/futures.py
```diff
@@ -155,6 +155,7 @@ def __init__(self, *, loop=None):
             self._source_traceback = traceback.extract_stack(sys._getframe(1))
 
     def _format_callbacks(self):
+        # Private method, do not rely on its existence.
```
gvanrossum: Would it make sense to rename to `__format_callbacks`? Or are there subclasses that call it?

asyncio/futures.py
```diff
@@ -360,25 +357,6 @@ def set_exception(self, exception):
 
     # Truly internal methods.
```
gvanrossum: This comment is out of date -- I don't believe `__iter__` is "truly internal".

asyncio/futures.py
```diff
@@ -20,7 +20,7 @@
 _FINISHED = 'FINISHED'
 
 Error = concurrent.futures._base.Error
-CancelledError = concurrent.futures.CancelledError
+CancelledError = concurrent.futures._base.CancelledError
```
gvanrossum: Curious why this was needed?

asyncio/futures.py
```diff
@@ -405,6 +390,26 @@ def _set_concurrent_future_state(concurrent, source):
         concurrent.set_result(result)
 
 
+def _apply_future_state(source, dest):
```
gvanrossum: I think this would be better off named `_copy_future_state()`.

gvanrossum:
Harmless, but suggesting some changes.

asyncio/futures.py
1st1: Yes, this is a good idea. Nothing is using this method besides the base Future class.

asyncio/futures.py
1st1: Agree

asyncio/futures.py
1st1: OK!

asyncio/futures.py
1st1: I think I've added it when I was debugging #287. I'll remove the change.

gvanrossum:
Ping me when you want me to have another look. (GitHub's PR code review confuses me because it's so different from the tools I've been using for 10 years.)

asyncio/futures.py
```diff
@@ -174,6 +175,7 @@ def format_cb(callback):
         return 'cb=[%s]' % cb
 
     def _repr_info(self):
+        # Private method, do not rely on its existence.
```
gvanrossum: This and the private methods below can also be given a `__` prefix right?

1st1:
> Ping me when you want me to have another look. (GitHub's PR code review confuses me because it's so different from the tools I've been using for 10 years.)

I've just pushed this commit that should address your comments: https://github.com/python/asyncio/commit/16851f70646f5f8cea73163d823ee9e2486c3e53

If it looks good, I'll squash all three commits in one and commit the diff by hand.

asyncio/futures.py
1st1: It's a bit more complicated.  It's "private" in a sense that nobody outside of asyncio shouldn't rely on it.  In asyncio, it's overloaded in Task, and in one other place--something that we can address when/if we decide to rewrite Future (and maybe Task) in C.

gvanrossum: Oh, yeah. This is what we call "protected" -- subclasses can use/override. So just update the comment (presumably also for the others?).

1st1: > Oh, yeah. This is what we call "protected" -- subclasses can use/override

Yeah, but I don't want for **non-asyncio** subclasses to rely on this method. "protected" means (at least in c++/java/etc) that it's OK for subclasses to override and/or use.  Maybe, instead of "private" I should just put "do not use outside of asyncio" there?

gvanrossum: Oh, sounds good.

1st1:
Another commit -- https://github.com/1st1/asyncio/commit/f63a33959825022299752d70d192677a7ca8397f. I've fixed the terminology, and also added this comment to `Task._step` and `Task._wakeup`

1st1:
BTW, `Future._loop` is used quite frequently in tests and in asyncio code.  Should I add `Future.get_loop()`, or we can do it later?

gvanrossum:
Sorry for jerking you around, but looking at this version I think I would like to go back to single underscores. `__private` is just too much of a pain to deal with in the debugger. And I don't think the comments about not using these things outside asyncio are useful.

Also, what is the purpose of this PR again? Why can't your libuv Future class define `_copy_state()` and `_set_result_unless_cancelled()`?

gvanrossum:
I think you've used up your review quota for the day. :-(

On Mon, Nov 16, 2015 at 2:33 PM, Yury Selivanov <notifications@github.com>
wrote:

> BTW, Future._loop is used quite frequently in tests and in asyncio code.
> Should I add Future.get_loop(), or we can do it later?
>
> —
> <https://github.com/python/asyncio/pull/292#issuecomment-157192683>.
>



-- 

1st1:
> Also, what is the purpose of this PR again? Why can't your libuv Future class define _copy_state() and _set_result_unless_cancelled()?

I don't want to reimplement them. There is no need to those methods to be a part of the Future API.  I hope that we'll accelerate Futures and Tasks in 3.5.x, and I want it to be as painless as possible.

gvanrossum:
Please describe your proposed strategy in more detail. What's going to be
in the accelerator module, what remains in Python, and what's going to be
in the libuv-specific module?

1st1:
> Please describe your proposed strategy in more detail. What's going to be
in the accelerator module, what remains in Python, and what's going to be
in the libuv-specific module?

It would be great if we can merge this PR before 3.5.1, so that I can implement the Future in C without implementing not strictly necessary `_copy_state()` and `_set_result_unless_cancelled()`. The less C code I have, the easier it would be to thoroughly review and test it. There is no real reason for those methods to be defined in the Future class.

Sometime after 3.5.1, I plan to have a stable C Future implementation as part of the uvloop project.  To make things simpler, I hope that we'll be able to integrate that implementation directly into asyncio for everybody's benefit.  Ideally, `futures.py` can become something like this:

```python
try:
    from ._futures import Future
except ImportError:
    class Future:
        ...

def _copy_future_state(fut, other_fut):
    ...

def _set_result_unless_cancelled(fut, result):
    ...

def _chain_future(source, destination):
    ...

# etc
```

I still have to play with implementing `asyncio.Task` in C.  If it gives another visible performance boost, we might want to do that too.  Or we can simply derive `asyncio.Task` from Future implemented in C.  But this is something quite remote to plan for right now.

gvanrossum:
Well it all sound way too rushed just so you can show off some benchmark
numbers. Sure, the changes are very minor and very unlikely to break
anything, but I've lost confidence that this is the end. E.g. if you're so
keen on having the minimal amount of C code, wouldn't you want to avoid
having to write _repr_info() and _format_callbacks() in C too? Honestly, I
don't mind if you commit the changes for _copy_state and
_set_result_unless_cancelled() but I don't really expect that this will
make the 3.5.2 changes any easier to digest.

1st1:
> Well it all sound way too rushed just so you can show off some benchmark
numbers.

Benchmark numbers are great, but the reason why I'm investing a ton of time in asyncio is because we're building a database (based on PostgreSQL) that's built with it.  Python allows us to iterate faster.  I'm trying to improve things where it's possible, and Future object is one of the core building blocks.

> Sure, the changes are very minor and very unlikely to break
anything, but I've lost confidence that this is the end. 

I understand.  Things that I try to fix/improve are only visible when you're trying to implement a compatible loop/interfaces from the outside of asyncio (most of my recent PR were related to that).

> E.g. if you're so
keen on having the minimal amount of C code, wouldn't you want to avoid
having to write _repr_info() and _format_callbacks() in C too?

Those are only used (and usable) in `Future.__repr__`. Right now, I'm not even implementing `__repr__`.  Same thing about `_schedule_callbacks` -- there is little reason to call it for anybody. In contrast, helpers like `_copy_future_state` or `_set_result_unless_cancelled` can be handy, and thus may be used by someone, therefore I think it's better to keep those things separate.

gvanrossum:
Well, but an acceptable C Future should have a compatible `__repr__` (or at
least one providing the same level of insight and details).

Anyway, go ahead and commit the refactor of `_copy_state()` and
`_set_result_unless_cancelled()`, but I'm still skeptical about the road
that lies ahead. ;-)

1st1:
> Well, but an acceptable C Future should have a compatible `__repr__` (or at
least one providing the same level of insight and details).

Absolutely.

> Anyway, go ahead and commit the refactor of `_copy_state()` and
`_set_result_unless_cancelled()`, but I'm still skeptical about the road
that lies ahead. ;-)

Merged (without "private" comments).

gvanrossum:
Thanks for your understanding! And good luck with libuv -- it sounds really
cool (and it was one of the things PEP 3156 explicitly wanted to encourage).

On Tue, Nov 17, 2015 at 9:15 AM, Yury Selivanov <notifications@github.com>
wrote:

> Well, but an acceptable C Future should have a compatible __repr__ (or at
> least one providing the same level of insight and details).
>
> Absolutely.
>
> Anyway, go ahead and commit the refactor of _copy_state() and
> _set_result_unless_cancelled(), but I'm still skeptical about the road
> that lies ahead. ;-)
>
> Merged (without "private" comments).
>
> —
> <https://github.com/python/asyncio/pull/292#issuecomment-157440242>.
>



-- 

Issue #288, opened 2015-11-12
Script freezes on await coroutine cancelled by wait_for if this coroutine contains any await inside it's CancelledError handling


germn:
```python
import asyncio


async def some_func():
    try:
        await asyncio.sleep(2)
    except asyncio.CancelledError:
        # await asyncio.sleep(1)
        print('cancelled')

async def main():
    coro = some_func()
    try:
        await asyncio.wait_for(coro, timeout=1)
    except asyncio.TimeoutError:
        await coro
        print('timeout')

if __name__ == "__main__":
    loop = asyncio.get_event_loop()
    loop.run_until_complete(main())
```
Output as expected:
```python
cancelled
timeout
```
But if you uncomment line inside ```some_func``` script will freeze (without printing). It can be avoided if we will use task instead of coroutine. But behavior with coroutine seems to be unclear: why everything works if line is commented and just freezes (without any error) if line is uncommented? If coroutine can't be awaited in this case shouldn't we get exception if we try to awaite it?

Windows 10, Python 3.5.0

gvanrossum:
I'm looking into this. I suspect it's something deep inside Task._step()... Not sure how easy it'll be to track it down.

gvanrossum:
So here's the problem (as you already indicated). The call `await_for(coro, timeout=1)` constructs a Task to wrap coro. But two lines below, when the `wait_for()` has timed out, the code uses `await coro`. I still have to dig deeper, but it appears that somehow the Task created by await_for() gets woken up when some_func() awaits the second sleep() call (the one commented out in the working version). IOW some_func() does catch the CancelledError but then `await` instantly gets cancelled again.

At this point my brain hurts and I still haven't quite figured out the complete sequence of events, nor an idea of how to fix it (if we can even fix it). I've got the feeling it's similar to the scenario warned against in the docstring for run_until_complete() -- two tasks wrapping the same coroutine.

Anyway, congrats on discovering this!

1st1:
I think I know what's going on here:

```python
import asyncio

async def some_func():
    try:
        await asyncio.sleep(2)
    except asyncio.CancelledError:
        await asyncio.sleep(1)
        print('cancelled')

async def main():
    coro = some_func()
    try:
        await asyncio.wait_for(coro, timeout=1)
    except asyncio.TimeoutError:
        await coro
        print('timeout')

if __name__ == "__main__":
    loop = asyncio.get_event_loop()
    loop.run_until_complete(main())
```

[...]

PR #293, opened 2015-11-18
Error if awaiting in parallel on the same coroutine (in debug mode)

1st1:
This patch prevents coroutines to be awaited by more than one waiters simultaneously.  It affects only `async def` coroutines by patching `CoroWrapper.__await__`; I'm a bit hesitant to touch `CoroWrapper.__iter__`.

See issue #288 for details.

### DIFF
--- a/asyncio/coroutines.py
+++ b/asyncio/coroutines.py
@@ -140,7 +140,13 @@ class CoroWrapper:
 
     if compat.PY35:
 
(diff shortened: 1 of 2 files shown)


gvanrossum:
LGTM (I prefer this over modifying CPython).

1st1:
Merged! Thanks for the review.

> I prefer this over modifying CPython

I agree.

PR #294, opened 2015-11-20
Drop "value" parameter from Task._step method.

1st1:


asyncio/tasks.py
```diff
@@ -231,12 +231,12 @@ def _step(self, value=None, exc=None):
         self._fut_waiter = None
 
         self.__class__._current_tasks[self._loop] = self
-        # Call either coro.throw(exc) or coro.send(value).
+        # Call either coro.throw(exc) or coro.send(None).
         try:
-            if exc is not None:
-                result = coro.throw(exc)
+            if exc is None:
+                result = coro.send(None)
```
gvanrossum: Shouldn't this be next(coro)?

1st1: No, `async def` coroutines don't have `__iter__` and `__next__`...

gvanrossum: That deserves a comment then!

Otherwise, LGTM.

1st1:
Merged by hand. @gvanrossum Thanks for the review! Here's the comment with an explanation of why not `next(coro)`: https://github.com/1st1/asyncio/commit/9b9ac203393c8fbd9ba575cf61cd52d76f8869cc

PR #295, opened 2015-11-20
Make tests that use TestCase.subTest work in Python 3.3

1st1:
Couldn't push recent updates to CPython repo -- a push hook ensuring proper indentation failed. I think that using `subTest` is a great idea, we just need to make it not fail in 3.3.


gvanrossum:
LGTM. Thanks!

PR #297, opened 2015-11-29
Add new StreamReader.readuntil() method.

socketpair:


asyncio/streams.py
```diff
@@ -425,34 +425,53 @@ def _wait_for_data(self, func_name):
             self._waiter = None
 
     @coroutine
-    def readline(self):
+    def readline(self, separator=b'\n'):
+        """
```
gvanrossum: The opening """ should be on the same line as the first word of the docstring.

asyncio/streams.py
```diff
@@ -425,34 +425,53 @@ def _wait_for_data(self, func_name):
...
+        Read chunk of data from the stream until `separator` is found.
+
+        On success, return chunk that ends with `separator`. If only partial line
+        can be read, return incomplete chunk without full `separator` at the end.
+
```
gvanrossum: Delete this blank line.

asyncio/streams.py
```diff
@@ -425,34 +425,53 @@ def _wait_for_data(self, func_name):
             self._waiter = None
 
     @coroutine
-    def readline(self):
+    def readline(self, separator=b'\n'):
+        """
+        Read chunk of data from the stream until `separator` is found.
+
+        On success, return chunk that ends with `separator`. If only partial line
```
gvanrossum: line -> chunk

asyncio/streams.py
```diff
@@ -425,34 +425,53 @@ def _wait_for_data(self, func_name):
...
+                    enough = True
+                    isep += seplen
+                elif seplen == 1:
+                    isep = len(self._buffer)
                 else:
-                    ichar += 1
-                    line.extend(self._buffer[:ichar])
-                    del self._buffer[:ichar]
-                    not_enough = False
+                    isep = 1 - seplen
```
gvanrossum: Can you make this `len(self._buffer) + 1 - seplen`? It's much clearer that way. (Also you can collapse the last two branches that way. :-)

asyncio/streams.py
```diff
@@ -425,34 +425,53 @@ def _wait_for_data(self, func_name):
...
+
+                line.extend(self._buffer[:isep])
+                del self._buffer[:isep]
 
                 if len(line) > self._limit:
+                    # self._buffer[0:0] = line
```
gvanrossum: Let's leave this out. If we were to push back the line we'd never make progress. (Leaving a potential partial separator in the buffer feels fine -- how much data gets dropped depend on exactly where the buffer boundaries are anyways, and that variability doesn't bother me.)

asyncio/streams.py
socketpair: fixed

asyncio/streams.py
socketpair: fixed

asyncio/streams.py
socketpair: fixed

asyncio/streams.py
socketpair: fixed

asyncio/streams.py
socketpair: Did not understand what you say :(. Will try to describe what I mean.

If we uncomment that, data in input buffer will never be lost. User should decide by its own what to do if too line gets beyond the limit. For example, he could try to re-read line with another limit, followed by dropping that data.

The common practice -- if some operation could not be done, raise exception and leave it in state as if operation has not been started. Leaving object in  unpredictable state - is a bad practice.

In current implementation, if last chunk (passed in `feed_data`) contain line end and some important data, that important data will be dropped unconditionally.

gvanrossum: But who says it's important data? It's most likely garbage, in the worst case an attack attempt. There is no API to change the limit on a stream already created, so code that were to catch the ValueError and try again will just be stuck in an infinite loop. Code that cares about this data should have started with a larger limit in the first place (the limit is set when the stream is created) or not use readline().

socketpair: OK, in that case we should raise exception for any attempt to read data after such `ValueError` has been fired. This will be clear solution. Should I add code for that ?

gvanrossum: Why? I don't believe there's something wrong with the existing code, and I
don't think we need to fix this issue as part of the read-until-delimiter
PR.

socketpair: OK. Will possibly create another PR for that infuture.

Next question: I want to `raise IncompleteReadError(line, None)` if complete line can not be read due to EOF. What you are thinking about that ? this behaviour will bemuch like readexactly, so after that implemented (is is very easy) it will not be required from user to check `.endswith(separator)` in order to detect incomplete line read.

gvanrossum: That would have to be a separate method, similar to readexactly() vs.
read(). It is pretty well established that readline() returns a partial
final line.

Question: perhaps instead of mucking with readline(), would it be more
useful to expose some details of the implementation so you can implement
your own read-until-separator code efficiently (i.e. without introducing an
extra layer of buffering)?

gvanrossum:
Ping? Any updates?

socketpair:
@gvanrossum pong. There is nothing to change in that PR as I think. If not - please say what should be done.

asyncio/streams.py
socketpair: Since data arrive in some chunks of unknown size, buffering is required. Will see how that implemented in tornado framework.

socketpair: Also i'm still dont fully understand what you say :(. Expose some details ? I have shown. My own implementation? Did you mean create `def read_until(separator)` instead of patching `readline` ? I will be happy to introduce new semantics in new function, if you accept that.

gvanrossum:
Here's what I recommend: instead of adding an optional argument to readline(), leave readline() alone and add a new method readrecord() with a mandatory trailing separator. You can then change the policies of readrecord() to be more strict when it encounters premature EOF or a record that's too long (or you could ignore the limit -- after all read() and readexactly() ignore the limit too).

socketpair:
This is pleriminary fix. I don't know if I should change `readline` code ever. Now, all test passed and no code duplication. Also, I use offset instead of merging/slicing buffers - this is faster than previous implementation.

So, you should decide:
1. Should I leave original `readline` implementation or not
2. make a quick code-review + English-related fixes.

I'm happy to change anything.

socketpair:
@gvanrossum ping

gvanrossum:
Sorry, I've been alternatingly busy and sick. But I'll get to this
eventually!

On Wed, Dec 9, 2015 at 1:12 PM, Коренберг Марк <notifications@github.com>
wrote:

> @gvanrossum <https://github.com/gvanrossum> ping
>
> —
> <https://github.com/python/asyncio/pull/297#issuecomment-163393066>.
>



-- 

1st1:
Why `readrecord`? When I think about records I assume that they have constant sizes, or are encoded in `[header+body-len][body]` format.

Maybe `readuntil` would be a better name?

socketpair:
@1st1 agree. Tornado call it `read_until`. I wIll rename to same function, but without underscore, as you said.

1st1:
BTW, do we consume the separator? Should we add a bool flag to either leave it in the buffer or consume it?

socketpair:
@1st1 coukd you provide case when we should preserve separator ?

1st1:
Does `readrecord` in your current patch consume the separator (i.e. will it be left for future consumption in the internal `StreamReader` buffer)? The docstrting doesn't mention that.

> @1st1 could you provide case when we should preserve separator ?

I don't know any concrete use case for that, but I can imagine that being useful for separators more complex than `\r\n`.  But I think this is something we can always add later.

socketpair:
Yes, if complete record is read successfully, separator is removed both from buffer and from returned value.

In any case, it is guaranteed, that removed separator is the same as argument of the function, so data is not lost.

gvanrossum:
It should consume (and return) the separator, similar to readline(). (The
one difference is at EOF -- IIUC if readuntil() sees EOF before seeing a
full separator it should raise EOFError (and also if it sees EOF
immediately -- similar to readexactly()).

On Fri, Dec 11, 2015 at 9:48 AM, Yury Selivanov <notifications@github.com>
wrote:

> Does readrecord in your current patch consume the separator (i.e. will it
> be left for future consumption in the internal StreamReader buffer)? The
> docstrting doesn't mention that.
>
> @1st1 <https://github.com/1st1> could you provide case when we should
> preserve separator ?
>
> I don't know any concrete use case for that, but I can imagine that being
> useful for separators more complex than \r\n. But I think this is
> something we can always add later.
>
> —
> <https://github.com/python/asyncio/pull/297#issuecomment-164001905>.
>



-- 

gvanrossum:
Oh, I see, that's different from readline(). Do you have a reason for that?

On Fri, Dec 11, 2015 at 10:02 AM, Коренберг Марк <notifications@github.com>
wrote:

> Yes, if complete record is read successfully, separator is removed both
> from buffer and from returned value.
>
> In any case, it is guaranteed, that removed separator is the same as
> argument of the function, so data is not lost.
>
> —
> <https://github.com/python/asyncio/pull/297#issuecomment-164004882>.
>



-- 

socketpair:
@gvanrossum yes, I have reason. First, this function will be used to read message of unknown size from a stream. I.e multipart form data or end of http header or even end of http headers(`\r\n\r\n`). In all cases, separator is a some protocol defined delimiter, it have no meaning in data it service for.

Next, storing separator in returned value have meaning for readline api, since presense of newline at end of string is only way to detect if complete line is read. In our case, incomplete line reading is acknowledged by exception. the same for overlimited line. In other words, for my function, if something is returned, it is guaranteed that this is complete data.

Next, after reading of such chunk, users will validate, parse it or pass to struct.unpack. In both cases, user does not need separator and he will cut it. Even for logging it is required to cut newline.

So, my function consume separator, but not return it. Seems this is very suitable for all tasks. If not, please provide example. Yes, it is not traditional (tornado returns separator), but much more convenient. In extra case, user may manually append separator to returned value before next processing of data, since separator was passed as argument. But I'm consider this is very rare case or even impossible in real situations. I have experience in writing protocol implementations.

1st1:
@socketpair All good points.

OTOH, I still don't like this subtle difference between `readline` and `readuntil`. Which one of them returns the delimiter is something that I will constantly forget.  Removing the delimiter is a matter of doing a simple slice operation.  Tornado, by the way, doesn't strip the delimiter: http://www.tornadoweb.org/en/stable/iostream.html#tornado.iostream.BaseIOStream.read_until

socketpair:
@1st1 I have renamed function

socketpair:
@1st1 and @gvanrossum. Will you accept this new function with current separator behaviour?

asyncio/streams.py
```diff
@@ -3,6 +3,7 @@
 __all__ = ['StreamReader', 'StreamWriter', 'StreamReaderProtocol',
            'open_connection', 'start_server',
            'IncompleteReadError',
+           'LimitOverrun',
```
socketpair: @1st1 and @gvanrossum Is this exception's name good enough? I have problems with naming and with my english.

asyncio/streams.py
```diff
@@ -35,6 +36,11 @@ def __init__(self, partial, expected):
         self.partial = partial
         self.expected = expected
 
+class LimitOverrun(Exception):
```
gvanrossum: The name is fine, but perhaps it should inherit from RuntimeError? (@1st1 what do you think?)

asyncio/streams.py
```diff
@@ -361,7 +367,7 @@ def set_exception(self, exc):
                 waiter.set_exception(exc)
 
     def _wakeup_waiter(self):
-        """Wakeup read() or readline() function waiting for data or EOF."""
+        """Wakeup read()/readline()/readuntil() function waiting for data or EOF."""
```
gvanrossum: Maybe `read*()` instead of naming every function?

asyncio/streams.py
```diff
@@ -426,34 +437,107 @@ def _wait_for_data(self, func_name):
...
+            return e.partial
+        except LimitOverrun as e:
+            # Emulate old behaviour...
+            if self._buffer.startswith(sep, e.consumed):
+                del self._buffer[:e.consumed + seplen]
+            else:
+                self._buffer.clear()
+            self._maybe_resume_transport()
+            raise ValueError(e.args[0])
+        return line + b'\n'
```
gvanrossum: Use `sep` here too.

gvanrossum:
I am almost happy with this. Left a few comments. Thanks for persevering (and sorry for being so slow -- being alternatingly sick and busy is no fun).

asyncio/streams.py
socketpair: fixed

asyncio/streams.py
socketpair: fixed

socketpair:
@gvanrossum you should not apologize, everyone understands your workload. :) I'm happy that you personally review my code.

socketpair:
I have add docstrings also, please review anyone and simplify `read()` function. This is becaue I wants to verify that it matches description.

asyncio/streams.py
```diff
@@ -30,11 +31,18 @@ class IncompleteReadError(EOFError):
     - expected: total number of expected bytes
     """
     def __init__(self, partial, expected):
-        EOFError.__init__(self, "%s bytes read on a total of %s expected bytes"
+        # `expected` may be int or None
+        # FIXME: replacing `EOFError` with `super()` will break tests...
```
socketpair: Seems tests are wrong, but don't know how to fix.

gvanrossum: Did you leave `self` in when using `super()`? If I write this it works for me:
```python
        super().__init__("%s bytes read on a total of %r expected bytes"
                                % (len(partial), expected))
```

socketpair: Yes, you are right! I have fixed.

gvanrossum:
I want to have one last conversation. Can you give me some insight in your use case? I find the difference with readline() somewhat jarring -- readline() returns the separator, readuntil() strips it. How important is that behavior for your use case? The reason I am asking is symmetry between reading and writing: when I am *writing* these records, I have to include the separator. But when *reading* them, you strip them. This means that code that just passes these records through to another stream must be careful to add the separators back. Reading lines using readline() does not have this issue -- you read exactly what you write, and you can write exactly what you read.

(I don't have an issue with raising an exception on an incomplete final record -- that behavior matches readexactly().)

socketpair:
@gvanrossum All my insights are shown earlier (https://github.com/python/asyncio/pull/297#issuecomment-164139857).

Reading and writing have subtle difference. Yes, I fully understand disbalance between reading and writing that you mean. But on other hand, we have no function `writewithseparator` just because it is much simplier to write chunk of data with already included in that data. Reading until separator is not so easy, as you see.

Also, you say that user should be careful when reading and writing. If I change code that will not strip separator from returned line, user's code will look like:

```
def write_to(stream, data):
    stream.write(data + '\r\n')

def read_from(stream):
   data = stream.readuntil('\r\n')
   return data[:-2]
```

does not it ?

I just want to simplify life of potential users of that.

P.S. I'm not fan of twisted, but http://twisted.readthedocs.org/en/latest/core/howto/servers.html#helper-protocols states: `Note that the delimiter is not part of the line.`. But it also have symmetrical function sendLine...

Okay, you win. I will add parameter - return separator or not. By default -- yes.

socketpair:
Now, all participants should be satisfied :)

socketpair:
Also add `limit` to `readline`. This will not hurt. + add test on limit strictness as well as documented that.

asyncio/streams.py
```diff
@@ -477,26 +598,37 @@ def read(self, n=-1):
...
+
+        Returned value is not limited with limit, configured at stream creation.
+
+        If steram was paused, this function will automatically resume it if
+        needed.
+        """
         if n < 0:
             raise ValueError('readexactly size can not be less than zero')
 
+        if n == 0:
```
gvanrossum: This changes behavior, right? If an exception is set it won't be raised. I think it should be moved past the test for `self._exception`.

asyncio/streams.py
```diff
@@ -30,11 +31,17 @@ class IncompleteReadError(EOFError):
     - expected: total number of expected bytes
     """
     def __init__(self, partial, expected):
-        EOFError.__init__(self, "%s bytes read on a total of %s expected bytes"
+        # `expected` may be int or None
+        super().__init__("%d bytes read on a total of %r expected bytes"
                                 % (len(partial), expected))
```
gvanrossum: Fix indent.

asyncio/streams.py
```diff
@@ -425,44 +440,150 @@ def _wait_for_data(self, func_name):
...
             if self._eof:
-                break
+                line = self._buffer[:]
+                self._buffer.clear()
+                self._maybe_resume_transport()
+                raise IncompleteReadError(bytes(line), None)
 
-            if not_enough:
-                yield from self._wait_for_data('readline')
+            # _wait_for_data() will resume readinf if paused due to buffer overlimit
```
gvanrossum: What is readinf?

asyncio/streams.py
```diff
@@ -425,44 +440,150 @@ def _wait_for_data(self, func_name):
...
+        If n is zero, return empty bytes object immediatelly.
+
+        If n is positive, this function try to read `n` bytes, and may return
+        less or equal bytes than requested, but at least one byte. If EOF was
+        received before any byte is read, this function returns empty byte
+        object.
+
+        Returned value is not limited with limit, configured at stream creation.
+
+        If steram was paused, this function will automatically resume it if
```
gvanrossum: Typo: stream.

asyncio/streams.py
```diff
@@ -477,26 +598,37 @@ def read(self, n=-1):
...
+        """Read exactly `n` bytes. Raise an `IncompleteReadError` if EOF
+        is reached before `n` bytes can be read, the
+        `IncompleteReadError.partial` attribute of the exception contains the
+        partial read bytes.
+
+        if n is zero, return empty bytes object.
+
+        Returned value is not limited with limit, configured at stream creation.
+
+        If steram was paused, this function will automatically resume it if
```
gvanrossum: Typo: stream.

asyncio/streams.py
```diff
@@ -425,44 +440,150 @@ def _wait_for_data(self, func_name):
...
+    def readuntil(self, separator=b'\n', limit=None, withseparator=True):
+        """Read chunk of data from the stream until `separator` is found.
+
+        On success, data and its separator is removed from internal buffer
+        (i.e. consumed). Returned value depends on `withseparator` argument.
+        If True (by default), return chunk including separator at the end. Else,
+        return received data without separator at the end.
+
+        If limit is not specified (or None) default stream limit is used. Limit
+        mean maximal length of chunk that is returned not counting separator.
```
gvanrossum: mean -> means

asyncio/streams.py
```diff
@@ -425,44 +440,150 @@ def _wait_for_data(self, func_name):
...
+
+        If EOF occurs and complete separator still not found,
+        IncompleteReadError(<partial data>, None) will be raised and internal
+        buffer become empty. This partial data may contain partial separator.
```
gvanrossum: become -> becomes (or will become)

gvanrossum:
Good work! I have one substantial issue (the test in readexactly()) and a few typos.

asyncio/streams.py
socketpair: fixed

asyncio/streams.py
socketpair: fixed

asyncio/streams.py
socketpair: fixed

asyncio/streams.py
socketpair: fix in all occurences

asyncio/streams.py
socketpair: fixed

asyncio/streams.py
socketpair: fixed both

asyncio/streams.py
socketpair: @gvanrossum I can fix that if you want. Should I fix ?

asyncio/streams.py
```diff
@@ -418,6 +427,13 @@ def _wait_for_data(self, func_name):
             raise RuntimeError('%s() called while another coroutine is '
                                'already waiting for incoming data' % func_name)
 
+        assert not self._eof
```
socketpair: also add this assert. _wait_for_data should never be called if we at EOF.

asyncio/streams.py
```diff
@@ -378,6 +384,8 @@ def _maybe_resume_transport(self):
             self._transport.resume_reading()
 
     def feed_eof(self):
+        # fixme: uncommenting this fail tests :(
```
socketpair: Unfortunatelly, this is True. Don't know how to fix.

socketpair:
@gvanrossum 
>  substantial issue (the test in readexactly())

What that mean? All test are passed (according to travis and my own tests)

asyncio/streams.py
1st1: @gvanrossum `RuntimeError` is used in `_wait_for_data` method, which is in turn used by all read* methods.  I don't think we want `except RuntimeError` blocks to catch our new exception.

My only comment re `LimitOverrun` is to call it `LimitOverrunError`.

1st1: @socketpair Please add a docstring for LimitOverrun.

1st1:
Do we really need the `withseparator` parameter?  I'd vote for keeping asyncio code as simple as possible -- cutting off the separator is a matter of a simple slice operation.

BTW, did we agree to add `limit` parameters to `readuntil` and `readline`?  We already have a limit that is set when `StreamReader` is instantiated.  Why do we need another way of setting them?

Also, I think we should rename `separator` to `delimiter`.

asyncio/streams.py
```diff
@@ -378,6 +384,8 @@ def _maybe_resume_transport(self):
...
+        # assert not self._eof, 'feed_eof while EOF was already set'
```
gvanrossum: Just delete the assert and the comment, calling feed_eof() twice does no harm.

asyncio/streams.py
gvanrossum: OK, please implement Yury's suggestions.

gvanrossum:
> > substantial issue (the test in readexactly())

> What that mean?

Sorry, I meant moving the check around, which you did.

> Also, I think we should rename separator to delimiter.

Disagree. `os.linesep` is the name for the standard LF (or CRLF) line separator. So please keep separator.

Regarding the `limit` parameters, I agree with Yury they are unneeded -- if you want a specific limit you should set it when you construct the stream. So please remobe.

Regarding the `withseparator` parameter, I also agree with Yury, please remove it. If we really find it is needed a lot in the future we could eventually add it, but if we add it now we can never remove it even if we find it is not useful. At the moment I care more about keeping the library code small than about supporting every possible use case.

(If you really object strongly please explain your use case rather than proposing a compromise.)

asyncio/streams.py
socketpair: I think feeding eof twice mean bug difinitely, since this event may occur only once. Does not it? Just as clsing closed file is a bug.

socketpair:
It seems, that limit, configured at stream creation is good for case when we don't care about length while reading line. Something like default limit.

But sometimes, specific limit needed. Either bigger or smaller. Some protocols define different limits on some states. Say, http line with url may require limit of 200 kb, while http header line may required to limit with 16kb. And so on. Yes, we can choose maximal limit and set it as default. Why not to allow to set limits precisely ? 10k connections of malicious user may eat too much memory if we dont limit lines correctly.

About withseparator. Suppose code:

```
data = await readuntil('\r\n\r\n')
process(data[:-4])
```

or this:

```
sep = '\r\n\r\n'
data = await readuntil(sep)
process(data[:-len(sep)])
```

Is it nice ? and also it is not optimal with memory usage since returned value is not memoryview.

gvanrossum:
The limit is really only there to protect against attacks based on sending
really long lines (which has been used in real life against naive HTTP
servers). Application-level limits that vary depending on where you are in
the protocol should really be implemented at the application level.

Regarding the separator, unless you reveal your use case I really don't
care. To clarify: a use case is an *actual* use you intend to make of the
feature, not a toy example of why it would be nice to have.

1st1:
@socketpair Mark, I've been thinking about this for some time, and here's my 2 cents (to add to Guido's comments):

**Limits**  

Right now the limit for `StreamReader` is used in the flow control; it is essentially a high-water mark.

> [..] Application-level limits that vary depending on where you are in
the protocol should really be implemented at the application level.

In my experience, when you implement a protocol with a help of `StreamReader`, you often have your application level kind of intervened with the protocol implementation level, since streams+coroutines is such a high level abstraction.  For instance, when parsing HTTP, you can call `await readline(limit=8*1024)` when you're reading headers, but have your high-water mark limit set to the default 16mb.  I think it'd be reasonable to discuss limits for `readline` and `readuntil`, but *not as part of this PR*.  Let's keep things minimal and easy to review.

**withseparator**  

> and also it is not optimal with memory usage since returned value is not memoryview. 

The internal buffering mechanism of `StreamReader` already copies data many times.  If you want a super efficient protocol implementation close to zero-copy, just subclass the `Protocol` class and don't use any high level abstractions, such as streams.  And nobody prevents you from wrapping the result of `readuntil` in a memoryview and make it efficient from that point, btw.

gvanrossum:
(Actually, the default limit is only 2**16 i.e. 64 KB. And I still think
it's better if the client writes
```
line = stream.readline()
if len(line) > 8192: raise ...
```
That's explicit. But otherwise I agree with Yury (obviously). Maybe instead
of adding limit parameters we could add a set_buffer_limit() method that
can change the limit dynamically?

On Mon, Dec 14, 2015 at 9:41 AM, Yury Selivanov <notifications@github.com>
wrote:

> @socketpair <https://github.com/socketpair> Mark, I've been thinking
> about this for some time, and here's my 2 cents (to add to Guido's
> comments):
>
> *Limits*
>
> Right now the limit for StreamReader is used in the flow control; it is
> essentially a high-water mark.
>
> [..] Application-level limits that vary depending on where you are in
> the protocol should really be implemented at the application level.
>
> In my experience, when you implement a protocol with a help of
> StreamReader, you often have your application level kind of intervened
> with the protocol implementation level, since streams+coroutines is such a
> high level abstraction. For instance, when parsing HTTP, you can call await
> readline(limit=8*1024) when you're reading headers, but have your
> high-water mark limit set to the default 16mb. I think it'd be reasonable
> to discuss limits for readline and readuntil, but *not as part of this PR*.
> Let's keep things minimal and easy to review.
>
> *withseparator*
>
> and also it is not optimal with memory usage since returned value is not
> memoryview.
>
> The internal buffering mechanism of StreamReader already copies data many
> times. If you want a super efficient protocol implementation close to
> zero-copy, just subclass the Protocol class and don't use any high level
> abstractions, such as streams. And nobody prevents you from wrapping the
> result of readuntil in a memoryview and make it efficient from that
> point, btw.
>
> —
> <https://github.com/python/asyncio/pull/297#issuecomment-164505704>.
>



-- 

socketpair:
I'm not agree with you. In order to have a chance to merge this PR I will remove `limit` from `readline` and also remove `withseparator`. :(

socketpair:
If everything OK, I can squash commits. Please review last commit.

1st1:
@socketpair LimitOverrun still has to be renames. Add a detailed docstring about it.

gvanrossum:
Yes, this too.

socketpair:
OOPS, forgot about rename. here it is: babee6b0ebc6b5b53d15afe752a993bfebd5d7bf

socketpair:
I have squashed everything to single commit

asyncio/streams.py
```diff
@@ -27,15 +28,28 @@ class IncompleteReadError(EOFError):
...
-        EOFError.__init__(self, "%s bytes read on a total of %s expected bytes"
-                                % (len(partial), expected))
+        super().__init__("%d bytes read on a total of %r expected bytes"
+                         % (len(partial), expected))
         self.partial = partial
         self.expected = expected
 
 
+class LimitOverrunError(Exception):
+    """
```
1st1: Could you please reformat this as

```python
class LimitOverrunError:
    """Reached buffer limit while looking for the separator.

    Attributes: ...
```

asyncio/streams.py
```diff
@@ -477,29 +596,40 @@ def read(self, n=-1):
...
+        # This will work right even if buffer is less than n bytes
+        data = bytes(self._buffer[:n])
+        del self._buffer[:n]
 
         self._maybe_resume_transport()
         return data
 
     @coroutine
     def readexactly(self, n):
+        """Read exactly `n` bytes. Raise an `IncompleteReadError` if EOF
```
1st1: The first line of the docstring has to be a complete one-line sentence followed by an empty newline.

1st1:
Some questions about LimitOverrunError:

@gvanrossum How does "LimitOverflowError" sound? Plays nice with the flow control terminology.  Or maybe plain old "BufferOverflowError" (although it might confuse some users)?

I see that there are two different string messages for this error:

1. "Separator is not found, and chunk exceed the limit"
2. "Separator is found, but chunk is longer than limit"

Maybe we better have two different exception classes instead?  @socketpair, would you want to recover differently depending on how `LimitOverrunError` was raised?  If not, then why do we even have the message argument?

asyncio/streams.py
```diff
@@ -426,43 +454,134 @@ def _wait_for_data(self, func_name):
...
+        If limit is reached, ValueError will be raised. In that case, if
+        newline was found, complete line including newline will be removed
+        from internal buffer. Else, internal buffer will be cleared. Limit is
+        compared against part of the line without newline.
+
+        If stream was paused, this function will automatically resume it if
+        needed.
+        """
+        sep = b'\n'
+        seplen = len(sep)
```
1st1: I don't think we need `sep` and `seplen` variables, just use `b'\n'` and `1` in the code.  That will actually improve the readability.

asyncio/streams.py
```diff
@@ -426,43 +454,134 @@ def _wait_for_data(self, func_name):
...
+        Configured stream limit is used to check result. Limit means maximal
+        length of chunk that can be returned, not counting the separator.
+
+        If EOF occurs and complete separator still not found,
+        IncompleteReadError(<partial data>, None) will be raised and internal
+        buffer becomes empty. This partial data may contain a partial separator.
+
+        If chunk cannot be read due to overlimit, LimitOverrunError will be raised
+        and data will be left in internal buffer, so it can be read again, in
+        some different way.
```
1st1: @gvanrossum @socketpair Why do we leave the data in the buffer?  Suppose one implements a MIME parser, and tries to read until they see a boundary.  Would it make sense for them to handle the `LimitOverrunError`, and instead of closing the connection, trying to recover with just reading the data?  Is there a realistic use case for this?

The reason I'm asking this, is because I don't like that `readline` and `readuntil` will behave differently.

asyncio/streams.py
```diff
@@ -426,43 +454,134 @@ def _wait_for_data(self, func_name):
...
-                else:
-                    ichar += 1
-                    line.extend(self._buffer[:ichar])
-                    del self._buffer[:ichar]
-                    not_enough = False
-
-                if len(line) > self._limit:
-                    self._maybe_resume_transport()
-                    raise ValueError('Line is too long')
+        # Consume whole buffer except last bytes, which length is
```
1st1: @socketpair It's hard for me to match your comments here with the code below.  

Could you please split this huge comment into smaller ones and move them closer to the relevant pieces of code?

1st1:
@socketpair there are a bunch of questions to be resolved before merging this. Please read my comments.  Sorry it takes so long for this PR, but it's a sizeable rewrite of the existing code + new methods.  We have to be careful here.  Thanks!

asyncio/streams.py
socketpair: `total number of bytes that was consumed` this i wrong. total number of bytes that should be consumed, and is guaranteed that over configured limit. When this exception is raised nothing is actually consumed.

Please help me how to write it in English correctly.

asyncio/streams.py
socketpair: In the past, it was. But I'm intentionally give names to that values, since it is not obvious what is `1` for example.

Anyways, if @gvanrossum approve, I will fix.

asyncio/streams.py
socketpair: In previous implementation there was configurable limit. now it has gone. So, user make a warning, and try to read with bigger limit, or, read by portions to skip that bad data...or, maybe read until another separator, trying to recover from error.

One of the most unwanted feature of readlint - is to leave stream in unpredicatable state when such error occur.

I can remove that odd (as you think) behaviour, but in that case I should add closing of connection right before raising such exception. (at least, disable any readings and simulate EOF).

asyncio/streams.py
socketpair: cannot split, since it comments single line of code. I can move it down and split by neline inside comment. OK?

asyncio/streams.py
socketpair: fixed

socketpair:
OK, I understand you in your requirements, this is OK. My PR int nodejs tooks about 4 months :) and succeeded after that (so, I have found one other megabug somewhat related to bug I found (assertion failed in C code))

asyncio/streams.py
1st1: > [..] and try to read with bigger limit, or, read by portions to skip that bad data...or, maybe read until another separator, trying to recover from error.

Unfortunately I haven't seen such an advanced code yet.  Usually the code just crashes, and the connection gets closed :)

> One of the most unwanted feature of readline - is to leave stream in unpredicatable state when such error occur.

Agree.  But... The `readline` method before your PR did exactly that?  Sorry, I'm a bit confused now.

asyncio/streams.py
```diff
@@ -426,43 +454,134 @@ def _wait_for_data(self, func_name):
...
+        """
+        sep = b'\n'
+        seplen = len(sep)
+        try:
+            line = yield from self.readuntil(sep)
+        except IncompleteReadError as e:
+            return e.partial
+        except LimitOverrunError as e:
+            if self._buffer.startswith(sep, e.consumed):
+                del self._buffer[:e.consumed + seplen]
```
1st1: Please add a comment explaining what's going on here.

asyncio/streams.py
1st1: Let me leave some comments below with suggestions on how we can annotate the lines.

asyncio/streams.py
socketpair: There was a comment, saying that this restore original behaviour of readline (see - I even did not touch original tests). Since "original behaviour" is unknown for people seeing that comment in future, I decide to remove comment.

asyncio/streams.py
```diff
@@ -426,43 +454,134 @@ def _wait_for_data(self, func_name):
...
+        #   analyze bytes of buffer that match partial separator.
+        #   This is slow and/or require FSM. For this case our
+        #   implementation is not optimal, since require rescanning
+        #   of data that is known to not belong to separator. In
+        #   real world, separator will not be so long to notice
+        #   performance problems. Even when reading MIME-encoded
+        #   messages :)
+
+        offset = 0
+        while True:
```
1st1:
```python
# Loop until we find `separator` in the buffer, exceed the buffer size,
# or an EOF is happened.
```

asyncio/streams.py
socketpair: Yes, readline before me just clears buffers (sometimes, depending on fact it found newline or not), and was not closing connection. You may see original, not so obvious code.

asyncio/streams.py
socketpair: You are welcome. I have some problems with my English, since it is not native language for me.

asyncio/streams.py
```diff
@@ -426,43 +454,134 @@ def _wait_for_data(self, func_name):
...
+            buflen = len(self._buffer)
+            if buflen - offset >= seplen:
```
1st1:
```python
# Check if we now have enough data in the buffer for `separator` to fit.
```

asyncio/streams.py
```diff
@@ -426,43 +454,134 @@ def _wait_for_data(self, func_name):
...
+                isep = self._buffer.find(separator, offset)
+                if isep != -1:
```
1st1:
```python
`separator` is in the buffer. `isep` will be used later to
retrieve the data.
```

asyncio/streams.py
```diff
@@ -426,43 +454,134 @@ def _wait_for_data(self, func_name):
...
+        #   buffer='abcdefghijklmnopqrS'. We may safely consume
+        #   everything except that last byte, but this require to
+        #   analyze bytes of buffer that match partial separator.
+        #   This is slow and/or require FSM. For this case our
+        #   implementation is not optimal, since require rescanning
+        #   of data that is known to not belong to separator. In
+        #   real world, separator will not be so long to notice
+        #   performance problems. Even when reading MIME-encoded
+        #   messages :)
+
```
1st1:
```python
# `offset` is the number of bytes from the beginning of the buffer
# where is no occurrence of `separator`.
```

asyncio/streams.py
1st1: > You are welcome. I have some problems with my English, since it is not native language for me.

NP, it's not my native language either.  I'm mostly asking for comments because I've troubles following the code, so I want to have almost every significant line of it annotated.

asyncio/streams.py
```diff
@@ -426,43 +454,134 @@ def _wait_for_data(self, func_name):
...
+                if isep != -1:
+                    break
+                # see upper comment for explanation
+                offset = buflen + 1 - seplen
+                if offset > self._limit:
+                    raise LimitOverrunError('Separator is not found, and chunk exceed the limit', offset)
 
             if self._eof:
-                break
+                chunk = bytes(self._buffer)
```
1st1: Should this `if` clause be the first `if` in this `while` loop?

asyncio/streams.py
```diff
@@ -426,43 +454,134 @@ def _wait_for_data(self, func_name):
...
+
+        offset = 0
+        while True:
+            buflen = len(self._buffer)
+            if buflen - offset >= seplen:
+                isep = self._buffer.find(separator, offset)
+                if isep != -1:
+                    break
+                # see upper comment for explanation
+                offset = buflen + 1 - seplen
```
1st1: You know what, instead of having the lengthy comment explaining the "+1" part here, maybe we can just have

```python
offset = buflen - seplen
```

?  It's way easier to understand (although I now follow your logic), and there is no way you can see any performance difference.

1st1: BTW, should it be

```python
offset = max(buflen - seplen, 0)
```

?

gvanrossum: I'll leave the rest of the review to Yury -- I trust him completely (even
his English -- note that English isn't my mother tongue either :-).

On Mon, Dec 14, 2015 at 1:55 PM, Yury Selivanov <notifications@github.com>
wrote:

> In asyncio/streams.py
> <https://github.com/python/asyncio/pull/297#discussion_r47564276>:
>
> > +        #   This is slow and/or require FSM. For this case our
> > +        #   implementation is not optimal, since require rescanning
> > +        #   of data that is known to not belong to separator. In
> > +        #   real world, separator will not be so long to notice
> > +        #   performance problems. Even when reading MIME-encoded
> > +        #   messages :)
> > +
> > +        offset = 0
> > +        while True:
> > +            buflen = len(self._buffer)
> > +            if buflen - offset >= seplen:
> > +                isep = self._buffer.find(separator, offset)
> > +                if isep != -1:
> > +                    break
> > +                # see upper comment for explanation
> > +                offset = buflen + 1 - seplen
>
> BTW, should it be
>
> offset = max(buflen - seplen, 0)
>
> ?
>
> —
> <https://github.com/python/asyncio/pull/297/files#r47564276>.
>



-- 

socketpair: 1. Algorythms should be as clear as possible. So I can just remove long comment, but prefer strict rules on searching pattern.
2. Maybe some day I implement Boyer Moore algorithm
3. Please provide case when `max(buflen - seplen, 0)` is required (I will add it to tests if required)

asyncio/streams.py
socketpair: No, since we have some data in buffer + eof flag triggered. So,  complete record may be here, and we should read it.

asyncio/streams.py
socketpair: fixed

asyncio/streams.py
socketpair: fixed

asyncio/streams.py
socketpair: fixed

asyncio/streams.py
socketpair: fixed

socketpair:
Seems, I have fixed everything requested, except long comment and `offset = buflen - seplen`. Will it be merged in current state, or it is required to fix something ?

I speak about REQUIREMENTS, i.e. changes, without of which you will not merge.

asyncio/streams.py
1st1: > Algorythms should be clear, I have checked all corner cases. So I can just remove long comment, but prefer strict rules.

That was **your** checking, and it's **your** preference.  However, I had to do my own checking, and **other** asyncio devs will do theirs, for as long as your code exists.  Reading your huge comment and studying all corner cases consumes a lot of our time for no actual performance gains.

> Maybe some day I implement Boyer Moore algorithm or something more appropriate (like KMP), which also allows to detect LimitOverrunError earlier on some data/separator combinations.

Feel free to make a PR when this day comes :)

> Please provide case when max(buflen - seplen, 0) is required (I will add it to tests if required)

You're right, we don't need it since the code is guarded with `if buflen - offset >= seplen:`.

asyncio/streams.py
1st1: Right.  I suggest adding a comment about this.

1st1:
> Seems, I have fixed everything requested, except long comment and offset = buflen - seplen. Will it be merged in current state, or it is required to fix something ?

I left you a comment re `offset = buflen - seplen`.

asyncio/streams.py
socketpair: fixed

socketpair:
Well, only question leaft is a big comment. Will you merge with that? Also, when I remove that `+1` tests hang (!). Discover why.

1st1:
> Well, only question leaft is a big comment. Will you merge with that? 

We'll merge the PR, it's only a matter of time to fix the remaining nits!

> Also, when I remove that +1 tests hang (!). I now discovering why.

That might be a good reason to keep that `+ 1`.  Or, maybe, it's a hint that there is another bug.  Please investigate.

gvanrossum:
Happy New Year everybody! There seems to be one open issue here (the investigation into why that +1 is needed). I will try to conduct a separate review. Then we can hopefully merge this. (Personally I have no problem with the big comment.)

gvanrossum:
I looked into the reason for the `+ 1` and it is correct. The reason the test hangs is simple: It sets the limit to 7, fills the buffer with 8 bytes (none of which are the separator, which has size 1), and then expects LimitOverrunError to be raised. Using `offset = bufsize + 1 - seplen`, offset becomes 8, which is > 7, so the exception is raised. But without `+ 1`, offset is only 7, which doesn't cause the exception. But the test neither sets EOF not adds more data, so then readuntil() simply hangs (there is no higher-level timeout in the test or in the stream code).

asyncio/streams.py
gvanrossum: I think the comment(s) should be restored. The only reason we have these two except clauses is for backwards compatibility, so we should state that. Otherwise in the future people may waste time trying to understand the reason or try to fix it and accidentally break code that depends on the old way. (Actually, the `return e.partial` is for compatibility with readline() methods on synchronous I/O classes; the backwards compatibility concern applies to deleting the consumed portion of the buffer only.)

asyncio/streams.py
gvanrossum: I think this is fine, even though it is an obscure border case. There is actually an alternative way to read that buffered data -- using plain read().

asyncio/streams.py
gvanrossum: (Alternatively, in the future people may legitimately decide to leave the excess data in the buffer for the LimitOverrunError case -- it helps that the only reason to not do that was backwards compatibility, not anything deeper.)

asyncio/streams.py
```diff
@@ -426,43 +454,150 @@ def _wait_for_data(self, func_name):
...
+                    raise LimitOverrunError('Separator is not found, and chunk exceed the limit', offset)
 
+            # Complete message (with full separator) may be present in buffer
+            # even when EOF flag is set. This may happen when the last chunk
+            # adds data which makes separator be found. That's why we check for
+            # EOF *ater* inspecting the buffer.
             if self._eof:
-                break
+                chunk = bytes(self._buffer)
+                self._buffer.clear()
```
gvanrossum: Maybe we should not clear the buffer here? The same argument applies -- there's another way to read the rejected data if an app really wants it. That may be better than stuffing it in the exception.

socketpair: @gvanrossum 

> Maybe we should not clear the buffer here?

1. It is compatibility with `readexactly`.

2. If `LimitOverrun` occurs, this may not mean some protocol violation (say, in HTTP there are no limits on length in various places). This exception is not "fatal", and we should preserve data in order to continue. For example, users of library may issue warning in logs and try to re-read with bigger limit.

3. If `IncompleteReadError` occurs, this mean protocol violation (or sudden disconnection). That mean requested operation can never complete. Yes, someone may want to re-read data in different way, but I can not provide real use case (as opposed to `LimitOverrun` case). Partially fetched record can not be used in a general, so this exception is "fatal". If someone wants to inspect what data still have been fetched - it have ability to. If I remove buffer cleanup, and user wants to get that partial data, he will required to call `.read()` in a loop in order to fetch everything unread, since we can not get amount of data in buffer (and call `readexactly`). Moreover, it is common to use this function to read data line-by-line, and so users should distinguish between empty and non-empty data after this exception happen. In other words, having data inside THIS exception is very suitable.

gvanrossum:
OK, you've convinced me. LGTM. Maybe Yury has some final remarks? Yury, do you want to merge?

1st1:
@gvanrossum Yes, I like this PR, and the new `readuntil` method. Please let me to experiment with the patch over the weekend and I can merge it on Monday.

@socketpair Would you be able to create an issue on bugs.python.org and draft a patch for Python documentation (update asyncio docs, whatsnew, etc)?

socketpair:
@1st1 
1. Thanks to spend time with my code :)
2. Yes, I able to. http://bugs.python.org/issue26050
3. Please point me which files I should update

asvetlov:
I believe it's https://github.com/python/cpython/blob/master/Doc/library/asyncio-stream.rst for doc update

1st1:
Mark, I've committed your changes in 47fafbb1125ca645495e92a544334f9effca9db0.

Thank you very much for your hard work on this PR. Please upload a draft patch with documentation for `readuntil` and `LimitOverrunError` to http://bugs.python.org/issue26050.

socketpair:
@1st1 Patch added http://bugs.python.org/file41595/asyncio-stream-doc-preliminary.patch

PR #302, opened 2015-12-09
Skip getaddrinfo if "host" is already an IP address.

ajdavis:
Response to this discussion:

https://groups.google.com/d/topic/python-tulip/-SFI8kkQEj4/discussion

### DIFF
--- a/asyncio/base_events.py
+++ b/asyncio/base_events.py
@@ -117,6 +117,28 @@ def _check_resolved_address(sock, address):
                              % (host, err))
 
 
+@functools.lru_cache(maxsize=1024)
+def _ipaddr_infos(host, port, family, type, proto):
+    # Try to skip getaddrinfo if "host" is already an IP address.
+    if proto not in {0, socket.IPPROTO_TCP, socket.IPPROTO_UDP}:
+        return None
+
+    if type == socket.SOCK_STREAM:
+        proto = socket.IPPROTO_TCP
+    elif type == socket.SOCK_DGRAM:
+        proto = socket.IPPROTO_UDP
+    else:
+        return None
+
+    try:
+        addr = ipaddress.ip_address(host)
+    except ValueError:
+        return None
+    else:
+        af = socket.AF_INET if addr.version == 4 else socket.AF_INET6
+        return [(af, type, proto, '', (host, port))]
+
+
 def _run_until_complete_cb(fut):
     exc = fut._exception
     if (isinstance(exc, BaseException)
@@ -535,36 +557,13 @@ class BaseEventLoop(events.AbstractEventLoop):
             logger.debug(msg)
         return addrinfo
 
-    @functools.lru_cache()
-    def _ipaddr_infos(self, host, port, family, type, proto):
-        """Try to skip getaddrinfo if "host" is already an IP address."""
-        if proto not in {0, socket.IPPROTO_TCP, socket.IPPROTO_UDP}:
-            return None
-
-        if type == socket.SOCK_STREAM:
-            proto = socket.IPPROTO_TCP
-        elif type == socket.SOCK_DGRAM:
-            proto = socket.IPPROTO_UDP
-        else:
-            return None
-
-        try:
-            addr = ipaddress.ip_address(host)
-        except ValueError:
-            return None
-        else:
-            af = socket.AF_INET if addr.version == 4 else socket.AF_INET6
-            infos = [(af, type, proto, '', (host, port))]
+    def getaddrinfo(self, host, port, *,
+                    family=0, type=0, proto=0, flags=0):
+        infos = _ipaddr_infos(host, port, family, type, proto)
+        if infos:
             fut = futures.Future(loop=self)
             fut.set_result(infos)
             return fut
-
-    def getaddrinfo(self, host, port, *,
-                    family=0, type=0, proto=0, flags=0):
-
-        infos = self._ipaddr_infos(host, port, family, type, proto)
-        if infos:
-            return infos
         elif self._debug:
             return self.run_in_executor(None, self._getaddrinfo_debug,
                                         host, port, family, type, proto, flags)
--- a/asyncio/base_events.py
+++ b/asyncio/base_events.py
@@ -72,6 +72,14 @@ def _format_pipe(fd):
         return repr(fd)
 
 
+# Linux's sock.type is a bitmask that can include extra info about socket.
+_SOCKET_TYPE_MASK = 0
+if hasattr(socket, 'SOCK_NONBLOCK'):
+    _SOCKET_TYPE_MASK |= socket.SOCK_NONBLOCK
+if hasattr(socket, 'SOCK_CLOEXEC'):
+    _SOCKET_TYPE_MASK |= socket.SOCK_CLOEXEC
+
+
 @functools.lru_cache(maxsize=1024)
 def _ipaddr_info(host, port, family, type, proto):
     # Try to skip getaddrinfo if "host" is already an IP. Since getaddrinfo
@@ -80,10 +88,10 @@ def _ipaddr_info(host, port, family, type, proto):
     if proto not in {0, socket.IPPROTO_TCP, socket.IPPROTO_UDP} or host is None:
         return None
 
-    # Linux's sock.type is a bitmask, can include SOCK_NONBLOCK or SOCK_CLOEXEC.
-    if type & socket.SOCK_STREAM:
+    type &= ~_SOCKET_TYPE_MASK
+    if type == socket.SOCK_STREAM:
         proto = socket.IPPROTO_TCP
-    elif type & socket.SOCK_DGRAM:
+    elif type == socket.SOCK_DGRAM:
         proto = socket.IPPROTO_UDP
     else:
         return None
--- a/asyncio/base_events.py
+++ b/asyncio/base_events.py
@@ -80,6 +80,7 @@ def _ipaddr_info(host, port, family, type, proto):
     if proto not in {0, socket.IPPROTO_TCP, socket.IPPROTO_UDP} or host is None:
         return None
 
+    # Linux's sock.type is a bitmask, can include SOCK_NONBLOCK or SOCK_CLOEXEC.
     if type & socket.SOCK_STREAM:
         proto = socket.IPPROTO_TCP
     elif type & socket.SOCK_DGRAM:
--- a/asyncio/base_events.py
+++ b/asyncio/base_events.py
@@ -23,6 +23,7 @@ import ipaddress
 import itertools
 import logging
 import os
+import re
 import socket
 import subprocess
 import threading
@@ -72,55 +73,16 @@ def _format_pipe(fd):
         return repr(fd)
 
 
-def _check_resolved_address(sock, address):
-    # Ensure that the address is already resolved to avoid the trap of hanging
-    # the entire event loop when the address requires doing a DNS lookup.
-    #
-    # getaddrinfo() is slow (around 10 us per call): this function should only
-    # be called in debug mode
-    family = sock.family
-
-    if family == socket.AF_INET:
-        host, port = address
-    elif family == socket.AF_INET6:
-        host, port = address[:2]
-    else:
-        return
-
-    # On Windows, socket.inet_pton() is only available since Python 3.4
-    if hasattr(socket, 'inet_pton'):
-        # getaddrinfo() is slow and has known issue: prefer inet_pton()
-        # if available
-        try:
-            socket.inet_pton(family, host)
-        except OSError as exc:
-            raise ValueError("address must be resolved (IP address), "
-                             "got host %r: %s"
-                             % (host, exc))
-    else:
-        # Use getaddrinfo(flags=AI_NUMERICHOST) to ensure that the address is
-        # already resolved.
-        type_mask = 0
-        if hasattr(socket, 'SOCK_NONBLOCK'):
-            type_mask |= socket.SOCK_NONBLOCK
-        if hasattr(socket, 'SOCK_CLOEXEC'):
-            type_mask |= socket.SOCK_CLOEXEC
-        try:
-            socket.getaddrinfo(host, port,
-                               family=family,
-                               type=(sock.type & ~type_mask),
-                               proto=sock.proto,
-                               flags=socket.AI_NUMERICHOST)
-        except socket.gaierror as err:
-            raise ValueError("address must be resolved (IP address), "
-                             "got host %r: %s"
-                             % (host, err))
+# Match "zone index" at end of IPv6 address, e.g. "fe80::1%lo0" for localhost.
+_zone_index_pat = re.compile(r'%\w+$')
 
 
 @functools.lru_cache(maxsize=1024)
 def _ipaddr_info(host, port, family, type, proto):
-    # Try to skip getaddrinfo if "host" is already an IP address.
-    if proto not in {0, socket.IPPROTO_TCP, socket.IPPROTO_UDP}:
+    # Try to skip getaddrinfo if "host" is already an IP. Since getaddrinfo is
+    # slow and, on some platforms, single-threaded, users might handle name
+    # resolution in their own code and pass in resolved IPs.
+    if proto not in {0, socket.IPPROTO_TCP, socket.IPPROTO_UDP} or host is None:
         return None
 
     if type == socket.SOCK_STREAM:
@@ -130,13 +92,63 @@ def _ipaddr_info(host, port, family, type, proto):
     else:
         return None
 
-    try:
-        addr = ipaddress.ip_address(host)
-    except ValueError:
-        return None
+    # On Windows, socket.inet_pton() is only available since Python 3.4.
+    if hasattr(socket, 'inet_pton'):
+        if family == socket.AF_UNSPEC:
+            try:
+                socket.inet_pton(socket.AF_INET, host)
+                af = socket.AF_INET
+            except OSError:
+                try:
+                    socket.inet_pton(socket.AF_INET6, host)
+                    af = socket.AF_INET6
+                except OSError:
+                    # "host" is not an IP address.
+                    return None
+        else:
+            try:
+                socket.inet_pton(family, host)
+                af = family
+            except OSError:
+                return None
     else:
-        af = socket.AF_INET if addr.version == 4 else socket.AF_INET6
-        return af, type, proto, '', (host, port)
+        try:
+            ipaddress.IPv4Address(host)
+            af = socket.AF_INET
+        except ValueError:
+            try:
+                ipaddress.IPv6Address(_zone_index_pat.sub('', host))
+                af = socket.AF_INET6
+            except ValueError:
+                return None
+
+        if family not in (socket.AF_UNSPEC, af):
+            # "host" is wrong IP version for "family".
+            return None
+
+    return af, type, proto, '', (host, port)
+
+
+def _check_resolved_address(sock, address):
+    # Ensure that the address is already resolved to avoid the trap of hanging
+    # the entire event loop when the address requires doing a DNS lookup.
+
+    if hasattr(socket, 'AF_UNIX') and sock.family == socket.AF_UNIX:
+        return
+
+    type_mask = 0
+    if hasattr(socket, 'SOCK_NONBLOCK'):
+        type_mask |= socket.SOCK_NONBLOCK
+    if hasattr(socket, 'SOCK_CLOEXEC'):
+        type_mask |= socket.SOCK_CLOEXEC
+
+    host, port = address[:2]
+    info = _ipaddr_info(host, port, sock.family, sock.type & ~type_mask,
+                        sock.proto)
+
+    if info is None:
+        raise ValueError("address must be resolved (IP address),"
+                         " got host %r" % host)
 
 
 def _run_until_complete_cb(fut):
@@ -560,7 +572,7 @@ class BaseEventLoop(events.AbstractEventLoop):
     def getaddrinfo(self, host, port, *,
                     family=0, type=0, proto=0, flags=0):
         info = _ipaddr_info(host, port, family, type, proto)
-        if info:
+        if info is not None:
             fut = futures.Future(loop=self)
             fut.set_result([info])
             return fut
--- a/asyncio/base_events.py
+++ b/asyncio/base_events.py
@@ -95,33 +95,29 @@ def _ipaddr_info(host, port, family, type, proto):
     # On Windows, socket.inet_pton() is only available since Python 3.4.
     if hasattr(socket, 'inet_pton'):
         if family == socket.AF_UNSPEC:
-            try:
-                socket.inet_pton(socket.AF_INET, host)
-                af = socket.AF_INET
-            except OSError:
-                try:
-                    socket.inet_pton(socket.AF_INET6, host)
-                    af = socket.AF_INET6
-                except OSError:
-                    # "host" is not an IP address.
-                    return None
+            afs = socket.AF_INET, socket.AF_INET6
         else:
+            afs = (family, )
+
+        for af in afs:
             try:
-                socket.inet_pton(family, host)
-                af = family
+                socket.inet_pton(af, host)
+                break
             except OSError:
-                return None
+                pass
+        else:
+            # "host" is not an IP address.
+            return None
     else:
         try:
-            ipaddress.IPv4Address(host)
-            af = socket.AF_INET
+            addr = ipaddress.IPv4Address(host)
         except ValueError:
             try:
-                ipaddress.IPv6Address(_zone_index_pat.sub('', host))
-                af = socket.AF_INET6
+                addr = ipaddress.IPv6Address(_zone_index_pat.sub('', host))
             except ValueError:
                 return None
 
+        af = socket.AF_INET if addr.version == 4 else socket.AF_INET6
         if family not in (socket.AF_UNSPEC, af):
             # "host" is wrong IP version for "family".
             return None
--- a/asyncio/base_events.py
+++ b/asyncio/base_events.py
@@ -118,7 +118,7 @@ def _check_resolved_address(sock, address):
 
 
 @functools.lru_cache(maxsize=1024)
-def _ipaddr_infos(host, port, family, type, proto):
+def _ipaddr_info(host, port, family, type, proto):
     # Try to skip getaddrinfo if "host" is already an IP address.
     if proto not in {0, socket.IPPROTO_TCP, socket.IPPROTO_UDP}:
         return None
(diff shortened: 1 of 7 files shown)


asyncio/base_events.py
```diff
@@ -533,9 +534,34 @@ def _getaddrinfo_debug(self, host, port, family, type, proto, flags):
...
 
+    def _ipaddr_infos(self, host, port, family, type, proto):
+        """Try to skip getaddrinfo if "host" is already an IP address."""
+        if proto == 0:
+            if type == socket.SOCK_STREAM:
+                proto = socket.IPPROTO_TCP
+            elif type == socket.SOCK_DGRAM:
+                proto = socket.IPPROTO_UDP
+            else:
+                return None
```
ajdavis: I'm unsure of this section, it reproduces more of getaddrinfo's own logic than I'm comfortable with.

1st1: It should be fine, though. AFAIK you can't really combine `SOCK_STREAM` with anything other than `IPPROTO_TCP`, same for UDP.  And they only make sense for `AF_INET` and `AF_INET6`.

asyncio/base_events.py
```diff
@@ -533,9 +534,34 @@ def _getaddrinfo_debug(self, host, port, family, type, proto, flags):
...
+
+        try:
+            addr = ipaddress.ip_address(host)
```
1st1: Can we put an LRU cache around this call?

asyncio/base_events.py
```diff
@@ -533,9 +534,34 @@ def _getaddrinfo_debug(self, host, port, family, type, proto, flags):
             logger.debug(msg)
         return addrinfo
 
+    def _ipaddr_infos(self, host, port, family, type, proto):
+        """Try to skip getaddrinfo if "host" is already an IP address."""
+        if proto == 0:
```
1st1: But what if `proto != 0` or `IPPROTO_IP`?  You should return if `proto not in {IPPROTO_TCP, IPPROTO_UDP}`

ajdavis: Good idea. But asyncio by default calls getaddrinfo with proto=0, and relies on the "type" argument to guide getaddrinfo's choice of protocol. So I think it should be `if proto not in {0, IPPROTO_TCP, IPPROTO_UDP}: return None`, right?

ajdavis:
I've added another patch now, thanks for your comments.

asyncio/base_events.py
```diff
@@ -533,9 +535,37 @@ def _getaddrinfo_debug(self, host, port, family, type, proto, flags):
             logger.debug(msg)
         return addrinfo
 
+    @functools.lru_cache()
```
1st1: Hm.  I don't want `lru_cache` internals to hold references to event loops.  How about we factor `_ipaddr_infos` method into a top-level function in `base_events.py`, returning `None` or a tuple.  This way we'll also share this cache between different loops.  I'd also bump the cache size to 1024 or something.

ajdavis: Great point; I've added another commit that moves this to module scope and increases the cache size.

1st1:
LGTM, but please wait for @gvanrossum's final approval. 

If you're going to commit this yourself, please squash all commits into one and merge by hand.  Don't forget to update the CPython repo!

ajdavis:
Thanks so much Yury. I'll squash and commit here once we hear from Guido, but I don't think I have permissions to update CPython.

asyncio/base_events.py
```diff
@@ -115,6 +117,28 @@ def _check_resolved_address(sock, address):
...
+    else:
+        return None
+
+    try:
+        addr = ipaddress.ip_address(host)
+    except ValueError:
+        return None
+    else:
+        af = socket.AF_INET if addr.version == 4 else socket.AF_INET6
+        return [(af, type, proto, '', (host, port))]
```
gvanrossum: Something makes me worry that a recipient of this function could mutate this list and mess up the cached value.

gvanrossum:
My suggested change should be simple enough. I have no other comments so you can commit with the suggested fix after either Yury or myself approves.

ajdavis:
New patch implements @gvanrossum's suggestion: _ipaddr_infos now returns a tuple and the call site calls list() on the result, to prevent inadvertent mutation of the value in the LRU cache.

ajdavis:
Done: _ipaddr_infos is now _ipaddr_info.

1st1:
@ajdavis Looks good!  Please merge the PR (and I'll sync CPython repo myself).

asyncio/base_events.py
```diff
@@ -535,7 +559,12 @@ def _getaddrinfo_debug(self, host, port, family, type, proto, flags):
 
     def getaddrinfo(self, host, port, *,
                     family=0, type=0, proto=0, flags=0):
-        if self._debug:
+        info = _ipaddr_info(host, port, family, type, proto)
+        if info:
```
gvanrossum: I'd use `is not None` here.

gvanrossum:
Cool to see the solution. LGTM (except maybe the `is not None`) issue. Whoopie!

haypo:
Sorry, I didn't read the history of this pull request.

_check_resolved_address() already has code to check if a string is a valid IPv4 or IPv6 address. I didn't remember why exactly, but I remember that we choose to not use the ipaddress module for this function. Recently, the call to _check_resolved_address() was made optional (only done in debug mode) for performance reason. But socket.inet_pton() is now used if available, instead of getaddrinfo(), for performance reason.

I would prefer to not use ipaddress *or* use ipaddress in both functions (_check_resolved_address() & _ipaddr_info()).

--

@functools.lru_cache(maxsize=1024)

Is it worth to use a cache on such simple function?

gvanrossum:
The reason why `_check_resolved_address()` exists is that otherwise people
might be calling `loop.sock_connect()` with a hostname, at which point the
`connect()` call in the `_socket` extension module will do the host lookup,
potentially hanging the thread until the DNS resolution comes back, and
also acquiring the getaddrinfo lock in certain cases (though it tries to
use `inet_pton()` if it can).

The most worrisome part of this code (from a quick glance) seems to be that
`getaddrinfo('localhost', 0)` may return IPv6 addresses with '%' in them,
which always cause a call to `getaddrinfo()` in `connect()`...

I do like the fact that the ipaddress module never calls getaddrinfo() --
it's pure text manipulation. But I don't know if it handles IPv6 address
like `'fe80::1%lo0'` (which is one of the ones returned for localhost on my
Mac).

1st1:
> _check_resolved_address() already has code to check if a string is a valid IPv4 or IPv6 address. I didn't remember why exactly, but I remember that we choose to not use the ipaddress module for this function. Recently, the call to _check_resolved_address() was made optional (only done in debug mode) for performance reason. But socket.inet_pton() is now used if available, instead of getaddrinfo(), for performance reason.

IIRC Guido didn't want asyncio to depend on ipaddress, which was a relatively new module in the stdlib back then.  I think it's OK now.

I like the idea of using `socket.inet_pton` though. Maybe we should rewrite the patch to use it when it's available, and fallback to ipaddress when it's not?

Here's what I propose:

1. Update `_ipaddr_info` to use `socket.inet_pton`.  If it's not available (in 3.3 on Windows), we fallback to use the `ipaddress` module.  (I don't like using `AI_NUMERICHOST` because we have to call `socket.getaddrinfo` again).

2. `_check_resolved_address` should be updated to use `_ipaddr_info`.

3. We can probably start calling  `_check_resolved_address` in release mode too.

> @functools.lru_cache(maxsize=1024)
> Is it worth to use a cache on such simple function?

The function is simple, the instantiation of `IPv4Address` or `IPv6Address` address isn't (LRU cache in 3.5 is written in C, it's quite fast to check the cache compared to parsing of addr string).

gvanrossum:
Sounds good.

On Sun, Dec 13, 2015 at 12:34 PM, Yury Selivanov <notifications@github.com>
wrote:

> _check_resolved_address() already has code to check if a string is a valid
> IPv4 or IPv6 address. I didn't remember why exactly, but I remember that we
> choose to not use the ipaddress module for this function. Recently, the
> call to _check_resolved_address() was made optional (only done in debug
> mode) for performance reason. But socket.inet_pton() is now used if
> available, instead of getaddrinfo(), for performance reason.
>
> IIRC Guido didn't want asyncio to depend on ipaddress, which was a
> relatively new module in the stdlib back then. I think it's OK now.
>
> I like the idea of using socket.inet_pton though. Maybe we should rewrite
> the patch to use it when it's available, and fallback to ipaddress when
> it's not?
>
> Here's what I propose:
>
>    1.
>
>    Update _ipaddr_info to use socket.inet_pton. If it's not available (in
>    3.3 on Windows), we fallback to use the ipaddress module. (I don't
>    like using AI_NUMERICHOST because we have to call socket.getaddrinfo
>    again).
>    2.
>
>    _check_resolved_address should be updated to use _ipaddr_info.
>    3.
>
>    We can probably start calling _check_resolved_address in release mode
>    too.
>
> @functools.lru_cache(maxsize=1024)
> Is it worth to use a cache on such simple function?
>
> The function is simple, the instantiation of IPv4Address or IPv6Address
> address isn't (LRU cache in 3.5 is written in C, it's quite fast to check
> the cache compared to parsing of addr string).
>
> —
> <https://github.com/python/asyncio/pull/302#issuecomment-164295976>.
>



-- 

1st1:
> I do like the fact that the ipaddress module never calls getaddrinfo() --
it's pure text manipulation. But I don't know if it handles IPv6 address
like `'fe80::1%lo0'` (which is one of the ones returned for localhost on my
Mac).

Interesting.  `'%lo0` is a zone index (more about it [here](https://en.wikipedia.org/wiki/IPv6_address#Link-local_addresses_and_zone_indices)), and it looks like `ipaddress` module doesn't support them.  We can probably use the following algorithm (as a fallback for `socket.inet_pton`):

1. Try `IPv4Address(address)`
2. Try `IPv6Address( re.sub(r'%\w+$', '', address) )`

@gvanrossum @haypo what do you think?

gvanrossum:
FWIW inet_pton() does support those (even though the man page on my Mac
doesn't mention it). Possibly the return value from getaddrinfo() varies by
system though? In my case I found this in my /etc/hosts, explaining why I
get it:
```
127.0.0.1    localhost
::1             localhost
fe80::1%lo0    localhost
```
I swear I didn't put that there, so it's probably due to some default Mac
initialization (or the Dropbox IT folks).

I don't know if it's acceptable to just ignore the % stuff if we don't have
inet_pton() but I feel strongly about not calling getaddrinfo() as a
callback (since its lock is what started this whole thing) so I'm willing
to give it a try and see if anyone screams.

--Guido


On Mon, Dec 14, 2015 at 10:06 AM, Yury Selivanov <notifications@github.com>
wrote:

> I do like the fact that the ipaddress module never calls getaddrinfo() --
> it's pure text manipulation. But I don't know if it handles IPv6 address
> like 'fe80::1%lo0' (which is one of the ones returned for localhost on my
> Mac).
>
> Interesting. '%lo0 is a zone index (more about it here
> <https://en.wikipedia.org/wiki/IPv6_address#Link-local_addresses_and_zone_indices>),
> and it looks like ipaddress module doesn't support them. We can probably
> use the following algorithm (as a fallback for socket.inet_pton):
>
>    1. Try IPv4Address(address)
>    2. Try IPv6Address( re.sub(r'%\w+$', '', address) )
>
> @gvanrossum <https://github.com/gvanrossum> @haypo
> <https://github.com/haypo> what do you think?
>
> —
> <https://github.com/python/asyncio/pull/302#issuecomment-164512893>.
>



-- 

ajdavis:
This is all clear, thanks - I'll send an updated PR in the next day or two.

1st1:
>  [..] but I feel strongly about not calling getaddrinfo() as a
fallback (since its lock is what started this whole thing) [..]

@ajdavis Please reflect this in a comment on why we use ipaddress module as a fallback.  I was re-reading the issue, and forgot why we shouldn't try to use getaddrinfo/AI_NUMERICHOST until I saw that Guido's message.

ajdavis:
I've implemented the plan @1st1 laid out. I commented in _ipaddr_info:

    # Try to skip getaddrinfo if "host" is already an IP. Since getaddrinfo is
    # slow and, on some platforms, single-threaded, users might handle name
    # resolution in their own code and pass in resolved IPs.

There's also a bit of refactoring in the tests, since many of them need the same Mock configuration.

asyncio/base_events.py
```diff
@@ -70,49 +73,78 @@ def _format_pipe(fd):
...
+    # the entire event loop when the address requires doing a DNS lookup.
+
+    if hasattr(socket, 'AF_UNIX') and sock.family == socket.AF_UNIX:
+        return
+
+    type_mask = 0
+    if hasattr(socket, 'SOCK_NONBLOCK'):
+        type_mask |= socket.SOCK_NONBLOCK
+    if hasattr(socket, 'SOCK_CLOEXEC'):
+        type_mask |= socket.SOCK_CLOEXEC
```
1st1: You don't need `type_mask` anymore.

ajdavis: Really? I use it to ensure the "type" parameter will be something _ipaddr_info understands, either SOCK_STREAM or SOCK_DGRAM. I thought this logic was in _check_resolved_address because some users might set the SOCK_NONBLOCK or SOCK_CLOEXEC flags on sockets before calling sock_connect, so we need to mask those bits out.

Another option is do this in _ipaddr_info:

    if type & socket.SOCK_STREAM

Instead of 

    if type == socket.SOCK_STREAM

1st1: >  I thought this logic was in _check_resolved_address because some users might set the SOCK_NONBLOCK or SOCK_CLOEXEC flags on sockets before calling sock_connect, so we need to mask those bits out.

No. These options were set in the old version of `_check_resolved_address` right before it called `socket.getaddrinfo`. They weren't even applied to the socket.

Second, `_check_resolved_address` is used to check live socket objects.  You shouldn't do any guessing at this point -- if `_ipaddr_info` refuses the addr, then be it so.

Third, `SOCK_NONBLOCK` and `SOCK_CLOEXEC` aren't even used in `_ipaddr_info` in any way.

asyncio/base_events.py
```diff
@@ -70,49 +73,78 @@ def _format_pipe(fd):
...
-                               proto=sock.proto,
-                               flags=socket.AI_NUMERICHOST)
-        except socket.gaierror as err:
-            raise ValueError("address must be resolved (IP address), "
-                             "got host %r: %s"
-                             % (host, err))
+            addr = ipaddress.IPv4Address(host)
+        except ValueError:
+            try:
+                addr = ipaddress.IPv6Address(_zone_index_pat.sub('', host))
```
1st1: I'm actually thinking that we should use `str.partition` here:

```python
# Cut off IPv6 zone index (appended to host, separated by '%' sign). 
# If it happens that by doing so we made an invalid address look valid, 
# the code will fail later on `sock.connect` or `sock.bind`.
addr = ipaddress.IPv6Address(host.partition('%')[0])
```

1st1:
We also need a unittest to make sure that `fe80::1%lo0`-like addresses are parsed by `_ipaddr_info` when `sock.inet_pton` is **or** isn't available.

asyncio/base_events.py
```diff
@@ -70,49 +73,78 @@ def _format_pipe(fd):
...
-                             % (host, exc))
+        if family == socket.AF_UNSPEC:
+            afs = socket.AF_INET, socket.AF_INET6
+        else:
+            afs = (family, )
+
+        for af in afs:
+            try:
+                socket.inet_pton(af, host)
+                break
```
1st1: I'd just return right here.

asyncio/base_events.py
```diff
@@ -70,49 +73,78 @@ def _format_pipe(fd):
...
+            except OSError:
+                pass
+        else:
+            # "host" is not an IP address.
+            return None
     else:
```
1st1: If we "return" instead of breaking, we can dedent this block of code.

asyncio/base_events.py
```diff
@@ -70,49 +73,78 @@ def _format_pipe(fd):
...
-    elif family == socket.AF_INET6:
-        host, port = address[:2]
+# Match "zone index" at end of IPv6 address, e.g. "fe80::1%lo0" for localhost.
+_zone_index_pat = re.compile(r'%\w+$')
+
+
+@functools.lru_cache(maxsize=1024)
+def _ipaddr_info(host, port, family, type, proto):
+    # Try to skip getaddrinfo if "host" is already an IP. Since getaddrinfo is
+    # slow and, on some platforms, single-threaded, users might handle name
```
1st1: It's not "slow", it's blocking and not thread-safe on many platforms.

asyncio/base_events.py
```diff
@@ -70,49 +73,78 @@ def _format_pipe(fd):
...
+            return None
     else:
-        # Use getaddrinfo(flags=AI_NUMERICHOST) to ensure that the address is
-        # already resolved.
-        type_mask = 0
-        if hasattr(socket, 'SOCK_NONBLOCK'):
-            type_mask |= socket.SOCK_NONBLOCK
-        if hasattr(socket, 'SOCK_CLOEXEC'):
-            type_mask |= socket.SOCK_CLOEXEC
         try:
```
1st1: We need a comment here explaining why we use `ipaddress` instead of `getaddrinfo / AI_NUMERICHOST` combination.  The reason is that while the call to getaddrinfo would be non-blocking, it would still involve acquiring a lock, since getaddrinfo isn't thread-safe.  And lock congestion is something that we're trying to avoid.

asyncio/base_events.py
```diff
@@ -70,49 +73,78 @@ def _format_pipe(fd):
...
+
+@functools.lru_cache(maxsize=1024)
+def _ipaddr_info(host, port, family, type, proto):
+    # Try to skip getaddrinfo if "host" is already an IP. Since getaddrinfo is
+    # slow and, on some platforms, single-threaded, users might handle name
+    # resolution in their own code and pass in resolved IPs.
+    if proto not in {0, socket.IPPROTO_TCP, socket.IPPROTO_UDP} or host is None:
+        return None
+
+    if type == socket.SOCK_STREAM:
```
1st1: Yes, you're right, we have to use `&` instead of `==` here.

ajdavis:
This has certainly been educational. Thanks for the simplifying suggestions; I've uploaded a new patch.

asyncio/base_events.py
```diff
@@ -70,49 +72,74 @@ def _format_pipe(fd):
...
+            except OSError:
+                pass
+        else:
+            # "host" is not an IP address.
+            return None
+
+    try:
+        addr = ipaddress.IPv4Address(host)
+    except ValueError:
+        host = host.partition('%')[0]
```
1st1: You shouldn't modify the `host` variable -- it should be returned as-is.

tests/test_base_events.py
```diff
@@ -32,6 +32,120 @@
...
+
+    @patch_socket
+    def test_ipaddr_info_no_inet_pton(self, m_socket):
+        del m_socket.inet_pton
+        self.test_ipaddr_info()
+
+    def test_check_resolved_address(self):
+        # sock = socket.socket(socket.AF_INET)
+        # base_events._check_resolved_address(sock, ('1.2.3.4', 1))
+        #
```
1st1: This commented-out code -- do we need it?

tests/test_base_events.py
```diff
@@ -32,6 +32,120 @@
...
+            (INET6, STREAM, TCP, '', ('::3', 1)),
+            base_events._ipaddr_info('::3', 1, UNSPEC, STREAM, TCP))
+
+        # IPv6 address with family IPv4.
+        self.assertIsNone(
+            base_events._ipaddr_info('::3', 1, INET, STREAM, TCP))
+
+        # IPv6 address with zone index.
+        self.assertEqual(
+            (INET6, STREAM, TCP, '', ('::3', 1)),
```
1st1: It should be `(INET6, STREAM, TCP, '', ('::3%lo0', 1))`

1st1:
> This has certainly been educational. Thanks for the simplifying suggestions; I've uploaded a new patch.

For me too :) I've left few more comments, but we're almost there.

asyncio/base_events.py
```diff
@@ -70,49 +72,74 @@ def _format_pipe(fd):
...
+            afs = socket.AF_INET, socket.AF_INET6
+        else:
+            afs = (family, )
+
+        for af in afs:
+            # Cut off IPv6 zone index (appended to host, separated by '%').
+            # If we happen to make an invalid address look valid, the code
+            # will fail later on sock.connect or sock.bind.
+            if af == socket.AF_INET6:
+                host = host.partition('%')[0]
```
1st1: This isn't necessary, `sockey.inet_pton` supports zone indexes.

ajdavis: On Mac, inet_pton accepts "::1%lo", and on Windows it accepts "::1%0". On Linux I can't get it to accept any address with a zone index, I don't think Linux's inet_pton supports zone indexes. I think we should keep stripping the zone index before passing to inet_pton, in order to reliably check whether the address is resolved or not. But you're right, I should return the original "host" parameter, not the one with zone index stripped.

1st1: Eh, I just checked and couldn't get it to parse zone index on Linux too :(  Then it's worth adding a comment there.

ajdavis:
New patch.

asyncio/base_events.py
```diff
@@ -70,49 +72,74 @@ def _format_pipe(fd):
...
 
+@functools.lru_cache(maxsize=1024)
+def _ipaddr_info(host, port, family, type, proto):
+    # Try to skip getaddrinfo if "host" is already an IP. Since getaddrinfo
+    # blocks on an exclusive lock on some platforms, users might handle name
+    # resolution in their own code and pass in resolved IPs.
+    if proto not in {0, socket.IPPROTO_TCP, socket.IPPROTO_UDP} or host is None:
+        return None
+
+    if type & socket.SOCK_STREAM:
```
1st1: Could you please add a comment that on Linux, `socket.type` can be a bit-mask.  That was something new to me, for instance.

1st1:
LGTM. @gvanrossum, please take a final look.

asyncio/base_events.py
```diff
@@ -70,49 +72,75 @@ def _format_pipe(fd):
...
-    # getaddrinfo() is slow (around 10 us per call): this function should only
-    # be called in debug mode
-    family = sock.family
-
-    if family == socket.AF_INET:
-        host, port = address
-    elif family == socket.AF_INET6:
-        host, port = address[:2]
-    else:
+    # Even though getaddrinfo with AI_NUMERICHOST would be non-blocking, it
```
1st1: Can you also move this comment ("Even though ...") to line 113, before we use the `ipaddress` module?  That will explain why we use it instead of `AI_NUMERICHOST`.

ajdavis: Done, new patch.

asyncio/base_events.py
```diff
@@ -70,49 +72,75 @@ def _format_pipe(fd):
...
+        proto = socket.IPPROTO_TCP
+    elif type & socket.SOCK_DGRAM:
+        proto = socket.IPPROTO_UDP
+    else:
+        return None
+
+    # On Windows, socket.inet_pton() is only available since Python 3.4.
+    if hasattr(socket, 'inet_pton'):
+        if family == socket.AF_UNSPEC:
+            afs = socket.AF_INET, socket.AF_INET6
```
gvanrossum: I personally would make this a list (and also on line 96 below). Tuples ought to be used for "record"-like structures, and only rarely for read-only sequences. Here there's no particular need for the sequence to be read-only, and the cost of creating a 2-item list vs. a 2-item tuple isn't really worth worrying about given all the other things we do here.

asyncio/base_events.py
```diff
@@ -70,49 +72,75 @@ def _format_pipe(fd):
...
+            # address look valid, we fail later in sock.connect or sock.bind.
+            try:
+                if af == socket.AF_INET6:
+                    socket.inet_pton(af, host.partition('%')[0])
+                else:
+                    socket.inet_pton(af, host)
+                return af, type, proto, '', (host, port)
+            except OSError:
+                pass
+        else:
```
gvanrossum: I don't really like this else. There is no break in the for-loop so you can just put a blank line here and dedent the return one level.

asyncio/base_events.py
```diff
@@ -70,49 +72,75 @@ def _format_pipe(fd):
...
 
+@functools.lru_cache(maxsize=1024)
+def _ipaddr_info(host, port, family, type, proto):
+    # Try to skip getaddrinfo if "host" is already an IP. Since getaddrinfo
+    # blocks on an exclusive lock on some platforms, users might handle name
+    # resolution in their own code and pass in resolved IPs.
+    if proto not in {0, socket.IPPROTO_TCP, socket.IPPROTO_UDP} or host is None:
+        return None
+
+    # Linux's sock.type is a bitmask, can include SOCK_NONBLOCK or SOCK_CLOEXEC.
```
gvanrossum: But do we know that the values are also bit masks on *other* platforms? And how common is the use case of masking those others?

asyncio/base_events.py
```diff
@@ -70,49 +72,75 @@ def _format_pipe(fd):
...
+                else:
+                    socket.inet_pton(af, host)
+                return af, type, proto, '', (host, port)
+            except OSError:
+                pass
+        else:
+            # "host" is not an IP address.
+            return None
+
+    # Even though getaddrinfo with AI_NUMERICHOST would be non-blocking, it
```
gvanrossum: Add an explicit comment explaining that this is the fallback code if inet_pton() is not available. Or use an else and indent the rest of the function.

gvanrossum:
Admirable work! I have a few style nits and one platform question.

asyncio/base_events.py
ajdavis: On Mac and Windows, "type" is just SOCK_STREAM, SOCK_DGRAM, or SOCK_RAW, there aren't any other bits you can combine with those integers. The flags SOCK_NONBLOCK and SOCK_CLOEXEC that you can "or" with the socket type are specific to Linux, as far as I can tell. In any case this code is portable. Are you just concerned that the comment is inaccurate, or is there something wrong with my actual code?

gvanrossum: Well, you're still making a cross-platform assumption that these three constants have no bits in common. What if SOCK_DGRAM was 2 and SOCK_STREAM was 3? I'd prefer not to make any assumptions about the numeric values of defined constants.

ajdavis: Oh I see. So I should put code like this back in:

https://github.com/python/asyncio/blob/master/asyncio/base_events.py#L101

1st1: @ajdavis Yeah, I think Guido is right.

ajdavis: Done, new patch.

gvanrossum:
Thanks! All looks good now. FWIW I'd prefer not to see your local revision history in the asyncio logs.

ajdavis:
Squashed and committed in https://github.com/python/asyncio/commit/39c135baf73762830148236da622787052efba19

1st1:
Pushed to CPython in https://hg.python.org/cpython/rev/51df143c098d and https://hg.python.org/cpython/rev/33c7980552a6

PR #303, opened 2015-12-10
Make Tasks check if Futures are attached to the same event loop

1st1:


gvanrossum:
LGTM, but let's wait until the OP gives it a thumbs-up. Thanks for the quick fix!

vxgmichel:
It's fine with me, thanks!

1st1:
BTW, there is no way to do the check if the future already has the result.  So the exception will only be raised for futures without results/exceptions set.

gvanrossum:
That's OK, those futures are harmless. Go ahead and merge.

--Guido (mobile)

1st1:
Merged. Thanks!

Issue #276, opened 2015-09-30
Let's go over all places where `except Exception` is used


gvanrossum:
...And replace it with `except BaseException` (or `except:`), making sure we understand the consequences. This would be better if a GeneratorExit is unexpectedly thrown into a coroutine (e.g. by calling its close() method), or if KeyboardInterrupt is used and then the loop is manually restarted.

PR #305, opened 2015-12-20
Handle BaseException in event loop.

mayfield:
Prior to this patch the event loop was only intended to handle Exception.  BaseException exceptions would break out of the event loop (mostly).  This patch treats all exceptions equally so application code can throw BaseException and its derivatives. 

Created this at Guido's request..

ref: https://groups.google.com/forum/#!topic/python-tulip/QK7oDGJwhXg

### DIFF
--- a/asyncio/base_events.py
+++ b/asyncio/base_events.py
@@ -116,12 +116,6 @@ def _check_resolved_address(sock, address):
 
 
 def _run_until_complete_cb(fut):
-    exc = fut._exception
-    if (isinstance(exc, BaseException)
-    and not isinstance(exc, Exception)):
-        # Issue #22429: run_forever() already finished, no need to
-        # stop it.
-        return
     fut._loop.stop()
 
 
@@ -321,15 +315,7 @@ class BaseEventLoop(events.AbstractEventLoop):
             future._log_destroy_pending = False
 
         future.add_done_callback(_run_until_complete_cb)
-        try:
-            self.run_forever()
-        except:
-            if new_task and future.done() and not future.cancelled():
-                # The coroutine raised a BaseException. Consume the exception
-                # to not log a warning, the caller doesn't have access to the
-                # local task.
-                future.exception()
-            raise
+        self.run_forever()
         future.remove_done_callback(_run_until_complete_cb)
         if not future.done():
             raise RuntimeError('Event loop stopped before Future completed.')
--- a/asyncio/base_subprocess.py
+++ b/asyncio/base_subprocess.py
@@ -183,7 +183,7 @@ class BaseSubprocessTransport(transports.SubprocessTransport):
             for callback, data in self._pending_calls:
                 loop.call_soon(callback, *data)
             self._pending_calls = None
-        except Exception as exc:
+        except BaseException as exc:
             if waiter is not None and not waiter.cancelled():
                 waiter.set_exception(exc)
         else:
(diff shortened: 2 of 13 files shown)


haypo:
Can you please elaborate the use case? Do you want to catch all base
exceptions? Or only KeyboardInterrupt?

Why not setting a handler for SIGINT signal to catch KeyboardInterrupt in
the regular adyncio control flow?

gvanrossum:
There was some discussion on the python-tulip list. See also
https://github.com/python/asyncio/issues/276 (which currently has no
motivation either).

IMO the main problem with the existing approach (catching Exception and
subclasses only) is that if you ever catch KeyboardInterrupt and then
attempt to clean up you are likely to find various internal asyncio data
structures corrupted, because of the inconsistent treatment of
BaseException. (It's not that the latter is never caught -- every 'finally'
clause catches it and there are some 'except BaseException' clauses in the
code as well.)

The signal solution is indeed nicer, because it ensures that interrupts are
treated as regular asyncio events, but it means you can't interrupt code
that's stuck in a tight CPU loop (e.g. while True: pass), and it requires
more sophistication from users.

I don't really think that catching BaseException everywhere we currently
catch Exception makes anything worse -- do you?

On Sun, Jan 3, 2016 at 2:37 AM, Victor Stinner <notifications@github.com>
wrote:

> Can you please elaborate the use case? Do you want to catch all base
> exceptions? Or only KeyboardInterrupt?
>
> Why not setting a handler for SIGINT signal to catch KeyboardInterrupt in
> the regular adyncio control flow?
>
> —
> <https://github.com/python/asyncio/pull/305#issuecomment-168486131>.
>



-- 

haypo:
2016-01-03 22:12 GMT+01:00 Guido van Rossum <notifications@github.com>:
> The signal solution is indeed nicer, because it ensures that interrupts are
> treated as regular asyncio events, but it means you can't interrupt code
> that's stuck in a tight CPU loop (e.g. while True: pass), and it requires
> more sophistication from users.

The C signal handler is always called immediatly, but the
KeyboardInterrupt is raised "later". I don't think that expensive
CPU-bound blocking call is interrupted by CTRL+c. And it would be a
really bad idea to call such function in the thread running an event
loop.

Victor

gvanrossum:
On Mon, Jan 4, 2016 at 12:45 AM, Victor Stinner <notifications@github.com>
wrote:

> 2016-01-03 22:12 GMT+01:00 Guido van Rossum <notifications@github.com>:
> > The signal solution is indeed nicer, because it ensures that interrupts
> are
> > treated as regular asyncio events, but it means you can't interrupt code
> > that's stuck in a tight CPU loop (e.g. while True: pass), and it requires
> > more sophistication from users.
>
> The C signal handler is always called immediatly, but the
> KeyboardInterrupt is raised "later". I don't think that expensive
> CPU-bound blocking call is interrupted by CTRL+c. And it would be a
> really bad idea to call such function in the thread running an event
> loop.
>

But such things happen all the time, by accident or while people are
learning, and most of the time the loop is written in Python, and the ^C
*will* interrupt it (though not on Windows IIRC).

-- 

kxepal:
> Can you please elaborate the use case? 

Another use case was found in Hypothesis testing library: https://github.com/DRMacIver/hypothesis/issues/292#issuecomment-189703510

1st1:
I'm going to close this PR. First it's fairly outdates. Second, we aren't sure that we have a problem with `BaseExceptions`. Seems that the corner case is actually handling `KeyboardInterrupt` correctly, but that will require a separate discussion and a different PR.

PR #309, opened 2016-01-09
Implement Timeout class

asvetlov:
@gvanrossum @1st1 please review

### DIFF
--- a/asyncio/tasks.py
+++ b/asyncio/tasks.py
@@ -1,9 +1,10 @@
 """Support for tasks, coroutines and the scheduler."""
 
-__all__ = ['Task', 'Timeout',
+__all__ = ['Task',
            'FIRST_COMPLETED', 'FIRST_EXCEPTION', 'ALL_COMPLETED',
            'wait', 'wait_for', 'as_completed', 'sleep', 'async',
            'gather', 'shield', 'ensure_future', 'run_coroutine_threadsafe',
+           'timeout',
            ]
 
 import concurrent.futures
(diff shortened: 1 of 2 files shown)


gvanrossum:
I'm curious -- perhaps the name should not be capitalized? Several classes in contextlib.py aren't capitalized either (closing redirect_stdout/stderr, suppress). OTOH ExitStack *is* capitalized. I wonder if that's because ExitStack actually has an API besides `__enter__` and `__exit__`. Since your Timeout has no additional public API, perhaps it should be named timeout()?

asvetlov:
Agree with lower cased name.
I can even add context manager support to `wait_for()`:
```
with asyncio.wait_for(1.5, loop=loop):
     yield from coro()
```
but really don't like this approach: it changes type of returned value depending on passed arguments.

gvanrossum:
Yeah, let's make it asyncio.timeout(), but let's not overload wait_for().
Maybe timeout() should be a function that returns a class instance (to
prevent people from trying to subclass or type-check it).

asvetlov:
Moved `Timeout` class to `timeout()` factory function.
I really doubt that people will use `Timeout` for subclassing or type checking but prefer stronger solution until we have a use case for introducing a class as public API.
We can get rid of the factory function later anyway.

1st1:
LGTM

gvanrossum:
Yes, looks good. Andrew, can you commit this yourself? It needs to be merged into Python 3.4, 3.5, 3.6. It would also be nice to have a Misc/NEWS item and some docs...

asvetlov:
On my way

asvetlov:
Done

gvanrossum:
Thanks!

AndreLouisCaron:
Was this ever merged back into the official `asyncio` module?  I'm using Python 3.5.1 and `from asyncio import timeout` raises an `ImportError`...

1st1:
@AndreLouisCaron `asyncio.timeout()` was supposed to be a new feature in `3.5.2`, but we decided that its design has to be tweaked. I think we'll finally have it in 3.6.

AndreLouisCaron:
OK thanks!

Out of curiosity, what needs to be changed?  One thing that strikes me as odd in the interface is that explicitly raising a `CancelledError` translates into a `TimeoutError` even if the task didn't time out.  I would expect that to go through (although I assume this will make the implementation trickier).

1st1:
There are some concerns about interoperability with other frameworks, read this email thread: https://mail.python.org/pipermail/async-sig/2016-June/000045.html

Maybe we'll also make it use the `async with` statement (instead of plain `with`).

As for CancelledError/TimeoutError -- I think that `asyncio.timeout()` behaved in exactly the same way as `asyncio.wait` -- throwing `CancelledError` to stop things, and propagating `TimeoutError` back to the calling coroutine.

AndreLouisCaron:
OK thanks!  I have some related questions, but I guess I'll take those to the mailing list.

coxley:
Absent from 3.6 -- are there any further plans for this?

1st1:
@coxley Thanks for bringing this issue up.  I'll be working on a PEP regarding new APIs for asyncio in 3.7 soon. Will likely include `asyncio.timeout` in it.

PR #310, opened 2016-01-11
Fix timeout docstring example

kxepal:
The `>>> ` prefix defines new expression, so the example actually had two of 
these: one `with` and one `yield`. This will cause two syntax errors if you run 
doctests: incomplete `with` block and bad indention for `yield`. 

To make it correct, for inner blocks need to use `...` prefix instead:

    >>> with asyncio.timeout(0.001):
    ...      yield from coro()

However, no else examples uses such notation, so  the example 
turned into more consistent state.

### DIFF
--- a/asyncio/tasks.py
+++ b/asyncio/tasks.py
@@ -743,8 +743,8 @@ def timeout(timeout, *, loop=None):
 
     For example:
 
-    >>> with asyncio.timeout(0.001):
-    >>>     yield from coro()
+        with asyncio.timeout(0.001):
+            yield from coro()
 
 
     timeout: timeout value in seconds


asvetlov:
LGTM, please merge

asvetlov:
Sorry, though it's aiohttp repo :)
Fixed by 7300dc14bbce515f85d44afb9edcb4ee16ee7f02, python 3.4-3.6 is updated.

gvanrossum:
Thanks for all the grunt work!

On Mon, Jan 11, 2016 at 2:28 AM, Andrew Svetlov <notifications@github.com>
wrote:

> Sorry, though it's aiohttp repo :)
> Fixed by 7300dc1
> <https://github.com/python/asyncio/commit/7300dc14bbce515f85d44afb9edcb4ee16ee7f02>,
> python 3.4-3.6 is updated.
>
> —
> <https://github.com/python/asyncio/pull/310#issuecomment-170501509>.
>



-- 

PR #313, opened 2016-01-15
Fix error message in asyncio.selector_events.

Paradisee:
Incorrect error message in the module asyncio.selector_events for the methods:
_SelectorSocketTransport.write
_SelectorSslTransport.write
_SelectorDatagramTransport.sendto.

The previous error was raising a Tuple:
TypeError: ('data argument must be byte-ish (%r)', <class 'str'>)

Patched:
TypeError: data argument must be a bytes-like object, not 'str'

http://bugs.python.org/issue26125


gvanrossum:
Merged manually. Thanks!

PR #322, opened 2016-02-21
Prevent StopIteration from being thrown into a Future

Rosuav:
Raising StopIteration inside a Future will either result in a spurious
return-like action, or a RuntimeError, depending on the Python version.

### DIFF
--- a/asyncio/futures.py
+++ b/asyncio/futures.py
@@ -342,7 +342,7 @@ class Future:
         if isinstance(exception, type):
             exception = exception()
         if type(exception) is StopIteration:
-            raise TypeError("StopException interacts badly with generators "
+            raise TypeError("StopIteration interacts badly with generators "
                             "and cannot be raised into a Future")
         self._exception = exception
         self._state = _FINISHED
--- a/asyncio/futures.py
+++ b/asyncio/futures.py
@@ -342,7 +342,8 @@ class Future:
         if isinstance(exception, type):
             exception = exception()
         if type(exception) is StopIteration:
-            raise TypeError("StopException interacts badly with generators and cannot be raised into a Future")
+            raise TypeError("StopException interacts badly with generators "
+                    "and cannot be raised into a Future")
         self._exception = exception
         self._state = _FINISHED
         self._schedule_callbacks()
--- a/asyncio/futures.py
+++ b/asyncio/futures.py
@@ -343,7 +343,7 @@ class Future:
             exception = exception()
         if type(exception) is StopIteration:
             raise TypeError("StopException interacts badly with generators "
-                    "and cannot be raised into a Future")
+                            "and cannot be raised into a Future")
         self._exception = exception
         self._state = _FINISHED
         self._schedule_callbacks()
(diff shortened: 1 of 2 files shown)


asyncio/futures.py
```diff
@@ -341,6 +341,8 @@ def set_exception(self, exception):
             raise InvalidStateError('{}: {!r}'.format(self._state, self))
         if isinstance(exception, type):
             exception = exception()
+        if isinstance(exception, StopIteration):
+            raise TypeError("StopException interacts badly with generators and cannot be raised into a Future")
```
asvetlov: I afraid the line is much longer than allowed 79 characters.

Rosuav: Would you prefer that I shorten the message, or arbitrarily break it into two lines?

asyncio/futures.py
```diff
@@ -341,6 +341,9 @@ def set_exception(self, exception):
             raise InvalidStateError('{}: {!r}'.format(self._state, self))
         if isinstance(exception, type):
             exception = exception()
+        if type(exception) is StopIteration:
+            raise TypeError("StopException interacts badly with generators "
+                    "and cannot be raised into a Future")
```
1st1: Please align the second line

Rosuav: Got any more style issues to hit me with? So far, there's been not a single comment about the actual _content_ of the patch, just fiddling with the styling.

1st1: Well, Guido suggested you the error message text, so I guess it's fine. One comment was about `isinstance`, and another about bad code style (pep8 and alignment as in the rest of asyncio codebase).  I'm not sure why you have this attitude.

tests/test_futures.py
```diff
@@ -76,6 +76,9 @@ def test_exception(self):
         f = asyncio.Future(loop=self.loop)
         self.assertRaises(asyncio.InvalidStateError, f.exception)
 
+        # StopIteration cannot be raised into a Future - CPython issue26221
+        self.assertRaises(TypeError, f.set_exception, StopIteration)
```
1st1: Please use `with self.assertRaisesRegex` and match the error message (at least partially).

asyncio/futures.py
Rosuav: Sorry. It's just that the breaking of the line sounded very much like what Raymond Hettinger said about line lengths - the most important reason to stick to 80-character lines isn't readability, nor tooling, it's to make sure you don't get whacked with the PEP 8 hammer.

tests/test_futures.py
Rosuav: Huh, good call on that one. I hadn't noticed that the error text was wrong.

gvanrossum:
LGTM. I'm not sure the optimization is really worth it but I do see some timing difference -- a simple command-line test says the type() version takes 90 ns while the isinstance() takes 135 ns, both with an argument that's a RuntimeError instance.

1st1:
Merged. Thanks, Chris!

Issue #314, opened 2016-01-25
connect_read_pipe causes "Exception ignored in:" on program exit


sbourdeauducq:
The following program prints "Exception ignored in:" when exiting. I am using Python 3.5.1. In my real application, the problem also occurs with a inherited pipe file descriptor from the parent process (wrapped with ``open(fd, "rb", 0)``) instead of ``sys.stdin``.
```python
import asyncio
import sys

loop = asyncio.get_event_loop()
reader = asyncio.StreamReader(loop=loop)
reader_protocol = asyncio.StreamReaderProtocol(reader, loop=loop)

# commenting this out makes the message disappear
loop.run_until_complete(loop.connect_read_pipe(lambda: reader_protocol, sys.stdin))

loop.close()

# to demonstrate message happens at exit
import time
time.sleep(5)
```

Martiusweb:
Hi,

Can you run your code with the environment variable ``PYTHONASYNCIODEBUG=1``:

    PYTHONASYNCIODEBUG=1 python your_test_case.py

It seems that since you don't close anything else than the loop, an exception happens in a __del__, and prevent the actual RessourceWarning to be displayed.

sbourdeauducq:
I had already tried ``PYTHONASYNCIODEBUG=1``, but since it didn't print anything more I thought I was setting it wrong. Same behavior on 3.5.1 and 3.4.4rc1.

sbourdeauducq:
Ping

gvanrossum:
What's the FULL exception? Is there anything after the "Exception ignored in:"? Usually there's more there.

Your  problem is closing the loop while it's still referenced, that's not a good idea. E.g. you may want to close the pipe first.

sbourdeauducq:
No, there isn't more. This bug causes Python to print exactly that: "Exception ignored in:"

sbourdeauducq:
Modifying the program to close the pipe AND THEN run a dummy coroutine makes the message disappear.

```python
import asyncio
import sys

loop = asyncio.get_event_loop()
reader = asyncio.StreamReader(loop=loop)
reader_protocol = asyncio.StreamReaderProtocol(reader, loop=loop)

r, w = loop.run_until_complete(loop.connect_read_pipe(lambda: reader_protocol, sys.stdin))
r.close()

@asyncio.coroutine
def workaround():
    yield
loop.run_until_complete(workaround())

loop.close()
```

vxgmichel:
[...]

PR #326, opened 2016-03-22
Fix unix pipe transport 'repr' methods (issue #314)

vxgmichel:
This PR is a fix proposal for the following issue:
- [#314 connect_read_pipe causes "Exception ignored in:" on program exit](https://github.com/python/asyncio/issues/314).

It includes a test method to reproduce the original issue.

### DIFF
--- a/asyncio/unix_events.py
+++ b/asyncio/unix_events.py
@@ -329,10 +329,10 @@ class _UnixReadPipeTransport(transports.ReadTransport):
         elif self._closing:
             info.append('closing')
         info.append('fd=%s' % self._fileno)
-        if self._pipe is not None and \
-           self._loop._selector is not None:
+        selector = getattr(self._loop, '_selector', None)
+        if self._pipe is not None and selector is not None:
             polling = selector_events._test_selector_event(
-                          self._loop._selector,
+                          selector,
                           self._fileno, selectors.EVENT_READ)
             if polling:
                 info.append('polling')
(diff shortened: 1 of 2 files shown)


asyncio/unix_events.py
```diff
@@ -329,7 +329,8 @@ def __repr__(self):
         elif self._closing:
             info.append('closing')
         info.append('fd=%s' % self._fileno)
-        if self._pipe is not None:
+        if self._pipe is not None and \
+           self._loop._selector is not None:
```
asvetlov: :-1: 
Third-party loops have no `._selector` attribute

asvetlov: :-1: 
Third-party loops have no `._selector` attribute

vxgmichel: True, but it is used line 335 anyway.

Martiusweb: Since the goal of your patch is to reduce the risk of those exceptions (AttributeError or TypeError) in repr, replacing `self._loop._selector` by `getattr(self._loop, '_selector', None)` makes sense and is sufficient.

vxgmichel: You're right, `self._loop._selector` is not used anywhere else in those classes.

Martiusweb:
LGTM, can we merge this?

1st1:
LGTM, I'll merge it later today.

1st1:
Pushed in 39b87620dbaf6278170c083401e2ae3e7601ff18

1st1:
Thanks, Vincent!

1st1:
Committed in CPython in https://hg.python.org/cpython/rev/d4facabab91a

vxgmichel:
Great! I guess the issue #314 should be closed now?

1st1:
Yep ;)

PR #328, opened 2016-03-29
Python Issue #26654: Inspect keyword arguments of functools.partial.

iceb0y:


asyncio/events.py
```diff
@@ -60,14 +62,13 @@ def _format_callback(func, args, suffix=''):
...
-    if args is not None:
-        func_repr += _format_args(args)
+    func_repr += _format_args_and_kwargs(args, kwargs)
     if suffix:
         func_repr += suffix
     return func_repr
 
 def _format_callback_source(func, args):
-    func_repr = _format_callback(func, args)
+    func_repr = _format_callback(func, args, {})
```
haypo: Pass None instead of {}, it's cheaper.

asyncio/events.py
```diff
@@ -35,23 +35,25 @@ def _get_function_source(func):
...
     Special case for a single parameter: ('hello',) is formatted as ('hello').
     """
     # use reprlib to limit the length of the output
-    args_repr = reprlib.repr(args)
-    if len(args) == 1 and args_repr.endswith(',)'):
-        args_repr = args_repr[:-2] + ')'
-    return args_repr
+    items = []
+    if args:
+        items.extend(map(reprlib.repr, args))
```
haypo: Maybe it's just me but i like list-comprehension:

items = [reprlib.repr(arg) for arg in args]

haypo:
CPython issue: https://bugs.python.org/issue26654

asyncio/events.py
iceboy-sjtu: It's for checking None.

asyncio/events.py
iceboy-sjtu: Done.

iceboy-sjtu:
Gentle ping. I think the PR is ready.

gvanrossum:
@hayo can you review and merge this?

hayo:
sorry, wrong hayo, you probably mean haypo :)

iceboy-sjtu:
@haypo

1st1:
@iceboy-sjtu Could you please sign CPython's contributor agreement?

iceboy-sjtu:
@1st1 Signed online. Thank you.

1st1:
@iceboy-sjtu Looking at this PR again -- a quick question: a lot of callable formatting functionality is already implemented as part of `inspect.signature` API. Would it be possible to just reuse that and drop all formatting logic from asyncio?

iceboy-sjtu:
@1st1 This is just a quick fix for the issue that the error message includes positional arguments but it does not include keyword arguments which may be misleading. I agree that the original code looks ugly, but let's leave it as is as it does not cause any problem.

btw There are two problems using `inspect.signature`. Firstly, it does not work on builtin functions such as `print`, `len`, `repr`. It also does not work on any composition that includes these "non-workable functions" by `functools.partial`.

```python
>>> inspect.signature(print)
ValueError: no signature found for builtin <built-in function print>
>>> inspect.signature(functools.partial(print, 0))
ValueError: no signature found for builtin <built-in function print>
```

Secondly, it does not recognize functools.partial.

```python
>>> str(inspect.signature(functools.partial(lambda x: x, 0)))
'()'
```

1st1:
> Firstly, it does not work on builtin functions such as print, len, repr

asyncio can't work with them either. Some builtin functions are just not introspectable yet.

> Secondly, it does not recognize functools.partial.

It does.  In your example it prints `()`, which means your callable that does not accept any arguments.  But I see the point now -- `inspect.signature` formats callable objects to show what arguments they can accept, whereas we want to format the function with all its bound arguments.  That means we can't use the signature API here...

asyncio/events.py
1st1: +1 for list-comprehension here.

iceboy-sjtu:
Acknowledged for `inspect.signature`. Seems good to merge?

asyncio/events.py
iceboy-sjtu: It's for checking `None`: `args` can be `None` in the new convention, because the added `kwargs` can be `None`, and it makes no sense that `args` cannot be `None`. Allowing `args` to be `None` simplifies the caller code, such as `event.py:55` and `event.py:65`.

With `args` being `None`, `[reprlib.repr(arg) for arg in args]` doesn't work.

1st1: > With args being None, [reprlib.repr(arg) for arg in args] doesn't work.

Line 45 checks if args is true, doesn't that help? I don't like `map` because it's much less readable.

iceboy-sjtu: Done. PTAL.

iceboy-sjtu:
Ping.

1st1:
Will merge this after beta1.

1st1:
Merged in b100957e7baf9482509631e504160d6c59509d26. Thanks a lot for fixing this!

PR #330, opened 2016-04-01
Add EPOLLEXCLUSIVE support to EpollSelector

methane:
EPOLLEXCLUSIVE was introduced in Linux 4.5.
It can be used to avoid thundering herd problem.

Reference:
http://man7.org/linux/man-pages/man2/epoll_ctl.2.html
https://lwn.net/Articles/633422/

### DIFF
--- a/asyncio/selectors.py
+++ b/asyncio/selectors.py
@@ -16,6 +16,9 @@ import sys
 EVENT_READ = (1 << 0)
 EVENT_WRITE = (1 << 1)
 
+# flag used by bitwise OR.  It may be ignored in unsupported platforms.
+EVENT_EXCLUSIVE = (1 << 2)
+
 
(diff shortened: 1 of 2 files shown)


1st1:
My understanding is that the `selectors` module isn't provisional, so this can only get into 3.6?

gvanrossum:
Right.

On Fri, May 13, 2016 at 11:41 AM, Yury Selivanov <notifications@github.com>
wrote:

> My understanding is that the selectors module isn't provisional, so this
> can only get into 3.6?
>
> —
> You are receiving this because you are subscribed to this thread.
> <https://github.com/python/asyncio/pull/330#issuecomment-219126741>
>



-- 

1st1:
@methane Please open an issue on Python bug tracker. I don't think this is related to asyncio anymore.

PR #333, opened 2016-04-20
better README format

ahxxm:
* fixed width for python binary command and file path

* display back slash

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -73,9 +73,9 @@ On Windows, things are a little more complicated.  Assume ``P`` is your
 Python binary (for example ``C:\Python33\python.exe``).
 
 You must first build the _overlapped.pyd extension and have it placed
-in the asyncio directory, as follows:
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -69,13 +69,13 @@ To run coverage (coverage package is required)::
 
     make coverage
 
-On Windows, things are a little more complicated.  Assume 'P' is your
-Python binary (for example C:\Python33\python.exe).
+On Windows, things are a little more complicated.  Assume ``P`` is your
+Python binary (for example ``C:\Python33\python.exe``).
 
 You must first build the _overlapped.pyd extension and have it placed
 in the asyncio directory, as follows:
```
rutsky: I think following paragraph should be formatted as code block (just use `::` as done above). Same with command line invocations examples below.

gvanrossum:
Thanks!

PR #339, opened 2016-05-02
Fix #338 os.writev() instead of os.write(b.join()) and sendmsg() instead of send()

### LINKED ISSUE #338: Serious performance loss (10 times) when NOT using .drain()
Linux Ubuntu 16.04, Python 3.5, standard asyncio library.

```python

#!/usr/bin/env python3.5

import asyncio
import os

async def test_read_arch(rs):
    while True:
        d = await rs.read(65536)
        if not d:
            break

async def test_write_arch(ws):
    b = b'x' * (1024 * 1024)
    for i in range(30):
        ws.write(b)
        # await ws.drain()  # <----------------- uncomment to speed program for 10 times (!)
    await ws.drain() # ensure everything is really written before closing.
    ws.close()

async def connect_write_pipe(file):
    transport, protocol = await  loop.connect_write_pipe(asyncio.streams.FlowControlMixin, file)
    stream_writer = asyncio.StreamWriter(transport, protocol, None, loop)
    return stream_writer, transport


async def connect_read_pipe(file):
    stream_reader = asyncio.StreamReader()
    transport, protocol = await loop.connect_read_pipe(lambda: asyncio.StreamReaderProtocol(stream_reader), file)
    return stream_reader, transport


async def amain():
    (r, w) = os.pipe()
    stream_reader, rtransport = await connect_read_pipe(open(r, 'rb'))
    stream_writer, wtransport = await connect_write_pipe(open(w, 'wb'))
    await asyncio.gather(
        test_write_arch(stream_writer),
        test_read_arch(stream_reader),
    )
    wtransport.close()
    rtransport.close()

def main():
    loop.run_until_complete(amain())


if __name__ == '__main__':
    main()
```

socketpair:


gvanrossum:
Please don't submit a PR that fails the tests and has typos in it. At the very least run it against your test program with and without drain().

socketpair:
@gvanrossum I have fix typo, and checked spped:

Before my patch:
without drain(): 4.5 sec
with drain():      0.2 sec

After my patch:
without drain(): 0.2 sec
with drain():      0.2 sec

gvanrossum:
I guess you have to mock os.writev too. Time to learn something!

gvanrossum:
BTW great news on the speed!

socketpair:
@gvanrossum 
Yes, I understand your fully. Have updated PR tests are fixed, but with ugly hack to overcome `@mock.patch('os.writev')` limitation....see comments

asyncio/unix_events.py
```diff
@@ -514,32 +514,41 @@ def write(self, data):
...
+        assert self._buffer, 'Data should not be empty'
 
-        self._buffer.clear()
         try:
-            n = os.write(self._fileno, data)
+            # !!!!!!!!!!!!!!!!!!!!! ONLY for WORKAROUNF BUG IN TESTS: 
+            # @mock.patch('os.writev') stores internal reference to _buffer, which is MODIFIED
+            # during this call. So, makeing deep copy :(
+            # !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+            n = os.writev(self._fileno, self._buffer.copy())
```
1st1: There must be a better way here.  Adding a copy just for convenient testing isn't a good idea.  Please also remove all exclamation marks, smiles, and fix typos.

asyncio/unix_events.py
```diff
@@ -514,32 +514,41 @@ def write(self, data):
...
     def _write_ready(self):
-        data = b''.join(self._buffer)
-        assert data, 'Data should not be empty'
+        # writing empty chunks to buffer is disallowed earlier
+        assert self._buffer, 'Data should not be empty'
 
-        self._buffer.clear()
         try:
-            n = os.write(self._fileno, data)
+            # !!!!!!!!!!!!!!!!!!!!! ONLY for WORKAROUNF BUG IN TESTS:
```
socketpair: dirty hack! How to eliminate this ?

asyncio/unix_events.py
socketpair: I consider you don't ever think to merge PR with such hack, so I added these comments to bring your attention. PR is not ready for merge. But I must point to bad places.

1st1: OK ;)  Please try to find a saner approach.  I'll myself try to take a look later.

socketpair: Has removed exclamation marks and other trash. But I stil don't know how to overcome that annoying bug in `unittest.mock`

gvanrossum:
I don't follow. Does the real `os.writev()` modify its buffer argument? Or the mock one?

socketpair:
@gvanrossum, in short code is:

```
buf = [1,2,3]
writev(buf) # < mock remembers REFERENCE to buf, not contents.
buf.pop()
```
Note, that `writev` was called with `[1,2,3]`, but mock see (in `assert`) that buf is `[2,3]`

asyncio/unix_events.py
```diff
@@ -514,32 +515,40 @@ def write(self, data):
...
+                # only part of chunk was written, so push unread part of it back to _buffer
+                self._buffer.appendleft(chunk[n:])
+                break
+
+        if self._buffer:
+            return
 
-            self._buffer.append(data)  # Try again later.
+        self._loop.remove_writer(self._fileno)
+        self._maybe_resume_protocol()  # May append to buffer.
```
socketpair: Seems another bug. in sockets, `self._maybe_resume_protocol()` is called before `if self._buffer`. So, who is right ?

gvanrossum: I would trust the socket code more. So I'd remove line 544-545 above and see how it fares.

1st1:
@socketpair Here's my attempt to tame the mock: https://github.com/1st1/asyncio/commit/5b8e2ddee4d6049877aefbe9464c48144939d077

socketpair:
@1st1 I think the most convenient way is to merge my PR in current state, and after that, merge you PR.

Is it suitable workflow ?

1st1:
@socketpair You can just incorporate my change into your PR. Please add a comment to `patched_os_writev` explaining why we need that craziness.

socketpair:
@1st1 @gvanrossum 

Sorry, please don't merge PR. I have checked performance. Using just expanding bytearray is much faster. C is not Python. :(

So please wait, I will revert everything and make bytearray in pipes.

Program used to benchmark:

```
#!/usr/bin/env python3.5

import asyncio
import os

async def handle_echo(rs, ws):
    for i in range(30):
        ws.write(b'x' * (10*1024*1024))
    print('writer all completely, drainig')
    await ws.drain()
    print('drainig complete')
    ws.close()


async def test_read_arch(rs, ws):
    s = 0
    while True:
        d = await rs.read(65536)
        if not d:
            break
        s += len(d)
    print('Read %d bytes completely' % s)
    ws.close()


async def amain():
    server = await asyncio.start_server(handle_echo, '127.0.0.1', 8888)
    (rs, ws) = await asyncio.open_connection('127.0.0.1', 8888)
    server.close()
    await server.wait_closed()

    await test_read_arch(rs, ws)

def main():
    loop.run_until_complete(amain())


if __name__ == '__main__':
    main()
```

1st1:
This is very strange. Using `bytearray` means more memory allocations and copy calls. From what I can see, `writev` implementation correctly uses buffer protocol, thus avoiding any extra memory work. What platform you're running your benchmark? What are the numbers?

gvanrossum:
Honestly I'm not that surprised that bytearray performs better. Resizing a
byte array is one of the fastest things you can do in CPython (because
`realloc()` handles the copy for you).

1st1:
Hi @socketpair! Any progress on this PR?

socketpair:
I'm online, will continue soon. I spent last week in another country, now I returned back to home, and today is a first working day at my job.

1st1:
Sure, no rush.  3.5.2 will happen relatively soon, would be nice to have this reviewed and merged before that.

socketpair:
great news! Now I know why bytearray was faster.

Now, my code is even faster than bytearray variant! since it true zerocopy! WOOHOO!

socketpair:
```
$ git checkout -f bytea 
$ time ./bench_pipes.py 
real	0m0.585s
user	0m0.336s
sys	0m0.244s
$ time ./bench_pipes.py 
real	0m0.535s
user	0m0.356s
sys	0m0.148s
$ time ./bench_pipes.py 
real	0m0.498s
user	0m0.344s
sys	0m0.152s
```
```
$ git checkout writev 
anya@mmarkk-imac:~/src/GH/asyncio$ time ./bench_pipes.py 
real	0m0.370s
user	0m0.284s
sys	0m0.084s
anya@mmarkk-imac:~/src/GH/asyncio$ time ./bench_pipes.py 
real	0m0.354s
user	0m0.280s
sys	0m0.068s
```

asyncio/selector_events.py
```diff
@@ -711,30 +711,44 @@ def write(self, data):
...
-                elif self._eof:
-                    self._sock.shutdown(socket.SHUT_WR)
+            return
+
+        while n:
+            chunk = self._buffer.popleft()
+            n -= len(chunk)
+            if n < 0:
+                # only part of chunk was written, so push unread part of it back to _buffer
+                # memoryview is required to eliminate memory copying while slicing.
```
socketpair: This was a cause of slow throughput

1st1:
Cool! Could you please fix the code style to 79 cars per line?

socketpair:
Now, I making commit for writelines optimisation. Please review current code.

asyncio/selector_events.py
```diff
@@ -617,7 +617,7 @@ def _call_connection_lost(self, exc):
                 self._server = None
 
     def get_write_buffer_size(self):
-        return len(self._buffer)
+        return sum(len(data) for data in self._buffer)
```
1st1: Maybe instead of running `sum` each time, we can have an internal `_buffer_length` attribute?

asyncio/selector_events.py
```diff
@@ -497,7 +497,7 @@ class _SelectorTransport(transports._FlowControlMixin,
 
     max_size = 256 * 1024  # Buffer size passed to recv().
 
-    _buffer_factory = bytearray  # Constructs initial value for self._buffer.
+    _buffer_factory = collections.deque  # Constructs initial value for self._buffer.
```
1st1: AFAIK, `deque` is a bit more memory hungry than `list`, since it operates on chunks of 64 pointers. Maybe this isn't an issue here?

asyncio/selector_events.py
```diff
@@ -711,30 +711,44 @@ def write(self, data):
...
-                if self._closing:
-                    self._call_connection_lost(None)
-                elif self._eof:
-                    self._sock.shutdown(socket.SHUT_WR)
+            return
+
+        while n:
+            chunk = self._buffer.popleft()
+            n -= len(chunk)
+            if n < 0:
```
1st1: Please make sure this branch is well tested.

asyncio/unix_events.py
```diff
@@ -520,32 +521,46 @@ def write(self, data):
...
+        # writing empty chunks to buffer is disallowed earlier
+        assert self._buffer, 'Data should not be empty'
 
-        self._buffer.clear()
         try:
-            n = os.write(self._fileno, data)
+            # list() is used only to workaround BUG in @mock.patch('os.writev').
+            # Mock stores internal reference to _buffer, which is then MODIFIED
+            # during this call. That's why we create separate superfluous copy.
+            n = os.writev(self._fileno, list(self._buffer))
```
1st1: This `list(self._buffer)` thing must go.

asyncio/unix_events.py
```diff
@@ -520,32 +521,46 @@ def write(self, data):
...
-            elif n > 0:
-                data = data[n:]
+            return
 
-            self._buffer.append(data)  # Try again later.
+        while n:
+            chunk = self._buffer.popleft()
+            n -= len(chunk)
+            if n < 0:
+                # only part of chunk was written, so push unread part of it back to _buffer
```
1st1: Same thing, this branch needs to be thoroughly tested.

bench_sockets.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env python3.5
+
+import asyncio
+import os
+
+async def handle_echo(rs, ws):
+    for i in range(30):
+        ws.write(b'x' * (10*1024*1024))
```
1st1: I'd also add a multiple small writes benchmark right here. If you add that, please use `TCP_NODELAY` to make small socket send ops faster and benchmark more stable.

1st1:
I've left some feedback.

asyncio/selector_events.py
socketpair: maybe, but I copy-pasted it from another place. Yes I can fix that (and perform benchmarking).

1st1: I suspect most of the time the `data` queue won't be large, so you really need to overwhelm the write buffer in your benchmarks to see the benefits of caching.

asyncio/selector_events.py
socketpair: 1. In that case, there will not be so much items in list.
2. deque was invented in order to speedup add/remove operations on both ends of a list. This is required for our operations.
3. Eating memory by deque is not issue at all. Buffer sizes are much bigger than memory used to maintain deque.

asyncio/unix_events.py
socketpair: Yes, I know. But tests become nightmare. Will fix a bit later.

asyncio/selector_events.py
socketpair: I need help in writing tests. It is not my strength side.

bench_sockets.py
socketpair: 1. OOPS. This file must not reach commit. Maybe put that file in `contrib` dir or so?
2. I will benchmark old and new code about many small writes.

asyncio/selector_events.py
1st1: > Eating memory by deque is not issue at all. Buffer sizes are much bigger than memory used to maintain deque.

Alright, let's use the `deque`.

asyncio/unix_events.py
1st1: In of the comments in this PR I sent you a link to my repo with tests being fixed. Please just use my approach.

socketpair:
Also, Python3.5 does not support `sendmmsg()`. If it will, I can also improve datagram transport

asyncio/selector_events.py
1st1: Well, you gotta try first :)

bench_sockets.py
1st1: `examples` dir is used for all kinds of code, I see no problem in this thing being added there too.

1st1:
> Also, Python3.5 does not support sendmmsg(). If it will, I can also improve datagram transport

You're welcome to write a patch. Or to at least create an issue (if it's not already there) on bugs.python.org.

asyncio/selector_events.py
socketpair: Sorry, I will not fix that. Code become too complex and error prone. Especially when exception occurs in some specific places. I tried.

bench_sockets.py
socketpair: many small writes suck. But win in original version. Will investigate why

socketpair: due to `sum(...)` :(

1st1: So we have to cache it.

socketpair: Fixed. But another problem: too many chunks in sendmsg lead to `OSError: [Errno 90] Message too long`. Will fix that.

socketpair: The operating system may set a limit (sysconf() value SC_IOV_MAX) on the number of buffers that can be used

1st1: What's the usual limit, or what order it usually is (10s, 100s, 1000s)?

socketpair: Fixed SC_IOV and buffer length caching.

asyncio/selector_events.py
socketpair: So, I have fixed anyway. I added many asserts.

asyncio/unix_events.py
socketpair: Yes, I will use. Wait.

socketpair:
oops. Sorry for the tests. Will fix

socketpair:
Okay, tests are fixed. I introduce new function, used only in tests. Is it right thing ?

asyncio/selector_events.py
```diff
@@ -28,6 +32,10 @@
 from .log import logger
 
 
+SC_IOV_MAX = os.sysconf(os.sysconf_names['SC_IOV_MAX'])
```
socketpair: not sure if it is available in other UNIX-like OS

asyncio/selector_events.py
```diff
@@ -787,6 +853,10 @@ def __init__(self, loop, rawsock, protocol, sslcontext, waiter=None,
             start_time = None
         self._on_handshake(start_time)
 
+    def _prefill_buffer(self, data):
```
socketpair: this function used only in tests. maybe rename ?

socketpair:
Added streams becnhmarking tool. Results:

Original asyncio:

```
bench_pipes: 5 buffers by 10.00 MB (with intermediate drains): 2.39 seconds (20.93 MB/sec)
bench_pipes: 5120 buffers by 0.01 MB (with intermediate drains): 0.17 seconds (295.10 MB/sec)
bench_sockets: 5 buffers by 10.00 MB (with intermediate drains): 0.18 seconds (275.94 MB/sec)
bench_sockets: 5120 buffers by 0.01 MB (with intermediate drains): 0.12 seconds (431.48 MB/sec)
bench_pipes: 5 buffers by 10.00 MB (without intermediate drains): 12.42 seconds (4.02 MB/sec)
bench_pipes: 5120 buffers by 0.01 MB (without intermediate drains): 13.65 seconds (3.66 MB/sec)
bench_sockets: 5 buffers by 10.00 MB (without intermediate drains): 0.20 seconds (250.48 MB/sec)
bench_sockets: 5120 buffers by 0.01 MB (without intermediate drains): 0.16 seconds (310.07 MB/sec)
```

Patched version:

```
bench_pipes: 5 buffers by 10.00 MB (with drain): 0.13 seconds (370.73 MB/sec)
bench_pipes: 5120 buffers by 0.01 MB (with drain): 0.14 seconds (347.25 MB/sec)
bench_sockets: 5 buffers by 10.00 MB (with drain): 0.11 seconds (457.13 MB/sec)
bench_sockets: 5120 buffers by 0.01 MB (with drain): 0.11 seconds (436.75 MB/sec)
bench_pipes: 5 buffers by 10.00 MB (without drain): 0.11 seconds (465.24 MB/sec)
bench_pipes: 5120 buffers by 0.01 MB (without drain): 0.15 seconds (335.80 MB/sec)
bench_sockets: 5 buffers by 10.00 MB (without drain): 0.11 seconds (465.32 MB/sec)
bench_sockets: 5120 buffers by 0.01 MB (without drain): 0.10 seconds (495.36 MB/sec)
```

asyncio/unix_events.py
socketpair: Finally, Have fixed test as @1st1 said.

asyncio/unix_events.py
socketpair: Fixed and make behaviour as in sockets.

asyncio/selector_events.py
socketpair: fixed

asyncio/selector_events.py
```diff
@@ -11,11 +11,15 @@
 import functools
 import socket
 import warnings
+import os
```
1st1: We sort imports alphabetically in asyncio.

asyncio/selector_events.py
```diff
@@ -11,11 +11,15 @@
...
+
 try:
     import ssl
 except ImportError:  # pragma: no cover
     ssl = None
 
+from itertools import islice
```
1st1: Just import `itertools`.

asyncio/selector_events.py
1st1: Will it fail with en exception? Or will it be `0`? Or?

asyncio/selector_events.py
```diff
@@ -519,11 +527,16 @@ def __init__(self, loop, sock, protocol, extra=None, server=None):
         self._protocol_connected = True
         self._server = server
         self._buffer = self._buffer_factory()
+        self._buffer_size = 0  # Cached count of bytes in buffer
         self._conn_lost = 0  # Set when call to connection_lost scheduled.
         self._closing = False  # Set when close() called.
         if self._server is not None:
             self._server._attach()
 
+    def _prefill_buffer(self, data):
```
1st1: You can just work with `_buffer` and `_buffer_size` directly in unittests.  Yes, it breaks the contract, but it's better than adding a tests-only method to the API.

asyncio/selector_events.py
```diff
@@ -704,43 +721,90 @@ def write(self, data):
...
 
         # Add it to the buffer.
-        self._buffer.extend(data)
+        self._buffer.append(data)
+        self._buffer_size += len(data)
         self._maybe_pause_protocol()
 
     def _write_ready(self):
         assert self._buffer, 'Data should not be empty'
+        assert self._buffer_size > 0, 'QWE!!! {} - {}'.format(self._buffer, self._buffer_size)
```
1st1: What is `QWE`?  Please combine the two asserts in one.

asyncio/selector_events.py
```diff
@@ -704,43 +721,90 @@ def write(self, data):
...
-            if not self._buffer:
-                self._loop.remove_writer(self._sock_fd)
-                if self._closing:
-                    self._call_connection_lost(None)
-                elif self._eof:
-                    self._sock.shutdown(socket.SHUT_WR)
+            return
+
+        assert n >= 0
+        assert n <= self._buffer_size
```
1st1: Do you really need these asserts? We better trust the `sendmsg` :)

asyncio/selector_events.py
1st1: As I said in other comment, just use `_buffer` directly in the tests (with a comment).

asyncio/selector_events.py
1st1: Or better, add a helper to `test_utils.py`.

tests/test_unix_events.py
```diff
@@ -35,6 +36,40 @@ def close_pipe_transport(transport):
...
 
 
+# unittest.mock store arguments by reference. So if argument is mutable
+# and changed after tested call, we can not check argument values correctly.
+# Example:
+# buf = [1,2,3]
+# writev(buf) # < mock remembers REFERENCE to buf, not contents.
+# buf.pop()
+# Note, that writev() was called with [1,2,3], but mock see (in assert) that
+# buf is [2,3]
```
1st1: Move this comment to a docstring, ident the example code for better readability and please don't use caps-lock ;)

1st1:
Some good progress on this PR, Mark!

AUTHORS
```diff
@@ -25,3 +25,4 @@ Victor Stinner <victor.stinner AT gmail.com>
 Vladimir Kryachko
 Yann Sionneau
 Yury Selivanov <yselivanov AT gmail.com>
+Коренберг Марк <socketpair AT gmail.com>
```
1st1: If possible, could you please write your name in english?

socketpair: Possible, but I want my name to be in Russian language, since I live in Russia. If English naming is strict for this project, I will rename.

asyncio/selector_events.py
```diff
@@ -1042,20 +1121,26 @@ def sendto(self, data, addr=None):
                                   'Fatal write error on datagram transport')
                 return
 
-        # Ensure that what we buffer is immutable.
-        self._buffer.append((bytes(data), addr))
```
socketpair: I don't know why this requirements appear. So I think twice, do some checks and decode to remove it in order to eliminate extra data copying.

asyncio/unix_events.py
```diff
@@ -475,21 +482,21 @@ def __repr__(self):
...
             logger.info("%r was closed by peer", self)
         if self._buffer:
+            assert self._buffer_size > 0
             self._close(BrokenPipeError())
         else:
             self._close()
 
     def write(self, data):
         assert isinstance(data, (bytes, bytearray, memoryview)), repr(data)
-        if isinstance(data, bytearray):
```
socketpair: I don't know why bytearray "converted" to memoryview. So, I decide to remove it.

asyncio/selector_events.py
socketpair: fixed

asyncio/selector_events.py
socketpair: fixed

asyncio/selector_events.py
socketpair: OK, will fix

asyncio/selector_events.py
socketpair: these size calculations are very error prone. I just insert asserts in all suspected places.

tests/test_unix_events.py
socketpair: fixed

asyncio/selector_events.py
socketpair: fixed

asyncio/selector_events.py
socketpair: fixed, but in slightly another way

socketpair:
So, left only tests on new functionality. Will make them soon, but not today.

Please review carefully. It is essential part of asyncio, and this PR is not so small. Hope I do right things.

asyncio/unix_events.py
socketpair: fixed

bench_sockets.py
socketpair: Usual limit of SC_IOV_MAX is 1024. On Linux kernel 2.2 it was 16, but we do not support this kernel anyway.

asyncio/selector_events.py
socketpair: Fixed. It checks that this constant is in `sysconf_names`. Also it asserts that result is > 0.

asyncio/selector_events.py
socketpair: Fixed

asyncio/selector_events.py
socketpair: @1st1  Trust `sendmsg`?
How about this: http://bugs.python.org/issue20113

AUTHORS
1st1: I think you're better to use English. Think about non-russian-speaking devs/users who won't be able to read/say your name.

asyncio/compat.py
```diff
@@ -1,17 +1,16 @@
 """Compatibility helpers for the different Python versions."""
 
+import os
 import sys
 
 PY34 = sys.version_info >= (3, 4)
 PY35 = sys.version_info >= (3, 5)
 
+x = os.sysconf_names.get('SC_IOV_MAX', None)
+if x is None:
```
1st1: I think if `x is None`, `SC_IOV_MAX` should be 1. That's the safest option.

asyncio/selector_events.py
1st1: Well, it was fixed, right? ;)  Please rewrite your asserts as `assert 0 <= n <= self._buffer_size`

asyncio/transports.py
```diff
@@ -93,11 +93,10 @@ def write(self, data):
...
         """Write a list (or any iterable) of data bytes to the transport.
 
-        The default implementation concatenates the arguments and
-        calls write() on the result.
+        The default implementation calls write() on each chunk of data.
         """
-        data = compat.flatten_list_bytes(list_of_data)
-        self.write(data)
+        for chunk in list_of_data:
+            self.write(chunk)
```
1st1: This is something we cannot do here, the ship has sailed. Please revert this line (or we have to consider enabling TCP_NODELAY by default for all connections, which is something that Golang does).

See my comment about this here: https://github.com/MagicStack/uvloop/blob/master/uvloop/handles/stream.pyx#L358

/cc @gvanrossum

asyncio/unix_events.py
```diff
@@ -513,39 +516,86 @@ def write(self, data):
...
+                # sendmsg() again and again until self._buffer is sent,
+                # or sendmsg() say that only part of requested data is sent.
+                # This will make code more complex. Instead, we assume that
+                # unsent part of self._buffer will be sent on next event loop
+                # iteration, even if one can be sent right now.
+                # According to benchmarking, this minor misoptimization does
+                # not hurt much. So I decide to keep code clear and simple.
+                #
+                # Unfortunatelly, Python 3.5 have a BUG - writev() does not
+                # accept generators. So, I wrap it to a list...
```
1st1: "So, I wrap it to a list..." -> "So, we wrap it in a list."

asyncio/unix_events.py
```diff
@@ -513,39 +516,86 @@ def write(self, data):
...
-                if not self._buffer and self._closing:
-                    self._loop.remove_reader(self._fileno)
-                    self._call_connection_lost(None)
-                return
-            elif n > 0:
-                data = data[n:]
+            return
+
+        assert n >= 0
+        assert n <= self._buffer_size
```
1st1: `assert 0 <= n <= self._buffer_size`

asyncio/unix_events.py
```diff
@@ -513,39 +516,86 @@ def write(self, data):
...
+            # Note, this is not length in bytes, it is count of chunks in buffer
+            if len(self._buffer) > compat.SC_IOV_MAX:
+                # In order to minimize syscalls used for IO, we should call
+                # sendmsg() again and again until self._buffer is sent,
+                # or sendmsg() say that only part of requested data is sent.
+                # This will make code more complex. Instead, we assume that
+                # unsent part of self._buffer will be sent on next event loop
+                # iteration, even if one can be sent right now.
+                # According to benchmarking, this minor misoptimization does
+                # not hurt much. So I decide to keep code clear and simple.
```
1st1: "So I decide to keep code clear and simple." > `/dev/null`

1st1:
Mark, I did a quick review. Will have to spend more time with the patch later. BTW, could you please upload the latest benchmark numbers?

asyncio/compat.py
methane: I think 16 is safe, according to http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/limits.h.html

1st1: That's POSIX. What about Windows?

methane: https://docs.python.org/3/library/os.html#os.writev
> Availability: Unix.

methane: I'm sorry, I hadn't realized that this patch uses sendmsg instead of writev.

https://msdn.microsoft.com/en-us/library/windows/desktop/ms741645(v=vs.85).aspx
MSDN doesn't remark about maximum number of dwBufferCount.

boost.asio uses:
```
# if defined (_WIN32_WINNT)
const int max_iov_len = 64;
# else
const int max_iov_len = 16;
# endif
```
https://github.com/boostorg/asio/blob/develop/include/boost/asio/detail/socket_types.hpp
https://github.com/boostorg/asio/commit/78d7795fbaf846d38dcecdae301e94f9df358740

So I believe max of `dwBufferCount` >= 64.

1st1:
> BTW, could you please upload the latest benchmark numbers?

Never mind, I've run them myself. Very impressive. The performance of asyncio is closer to uvloop with this PR (even faster for 5120 buffers, something I'll have to fix ;)

asyncio/compat.py
socketpair: 1. I assume that if that constant is missing than no limitation exists. Can someone confirm or refute that ? Maybe some practical tests on BSD, macosX, aix or so?
2. Well, I will change assertion that this constant must be >= 16.

socketpair: I tend to hardcode value 1024 when constant is missing.
I have checked: macos x and FreeBSD 10.3 gives us 1024.

Using 16 or even 1 lead to bad performance.

methane:
This pull request uses sendmsg, not writev.  Am I correct?
If so, how about changing title of this pull request?

socketpair:
@methane title is changed. Both functions are really fixed

asyncio/compat.py
1st1: Let's use 16, unless you want this PR to wait until Python 3.6. We cannot risk breaking asyncio on some platform in a bugfix Python release.

asyncio/transports.py
1st1: Actually, I think I'm wrong here. Since we're buffering, and still write data in one syscall, it should be fine. Keep the `for`.

1st1: Spent half a day profiling this stuff. See [1]. Replacing `write` with `writelines` (the one that calls write multiple times and uses zero-copy) slows things down 20-30%.

[1] https://github.com/MagicStack/vmbench/blob/master/servers/asyncio_http_server.py#L37

asyncio/unix_events.py
socketpair: fixed (in unix events and in selector events)

asyncio/unix_events.py
socketpair: Fixed all such asserts + all consecutive asserts created by me.

asyncio/unix_events.py
socketpair: fixed

asyncio/transports.py
socketpair: Can you provide code for benchmarking ?

1st1: I managed to get this working in uvloop -- i.e. `writelines` simply calls `write` in a loop for all buffers. However, to do that properly and without losing in performance, I had to reengineer how write buffers are emptied. Essentially, each transport in uvloop, when it needs to do a write op, simply stores the data in its private queue, and asks the loop to call it later. "Later" is after all callbacks are executed, but before the IO polling.

This would require a much deeper changes, which I wouldn't accept for 3.5.2. So for now, please, just join the buffers and call one `write` per `writelines`.  The difference is only visible when you have to join huge buffers (> 100 kb).

1st1: > Can you provide code for benchmarking ?

https://github.com/MagicStack/vmbench/blob/master/servers/asyncio_http_server.py#L37

Run this script with `--type=asyncio+httptools`. You can change `write(b''.join([...]))` to `writelines([..])`

asyncio/compat.py
socketpair: Fixed. This is OK, since it have no impact on modern operation systems.

asyncio/selector_events.py
socketpair: No, :) AFAIK we support Python 3.3, in some versions of which this bug reproduced.

asyncio/compat.py
gvanrossum: Just an observation -- if this change to asyncio requires changes to
uvloop, then that shows there's too much coupling between the two --
asyncio doesn't seem to have a sufficiently well-defined API that uvloop
can use (or this PR changes said API -- but I suspect there's no API
because none is documented).

AUTHORS
```diff
@@ -1,5 +1,5 @@
-A. Jesse Jiryu Davis <jesse AT mongodb.com>
 Aaron Griffith
+A. Jesse Jiryu Davis <jesse AT mongodb.com>
```
socketpair: Oh. I just sorted file using `sort`. Should I revert these lines?

asyncio/compat.py
1st1: @gvanrossum 

> Just an observation -- if this change to asyncio requires changes to
uvloop[..]

No, not really.  But I've used some of the ideas in this PR to benchmark and tune uvloop performance.

>  [..] then that shows there's too much coupling between the two --
asyncio doesn't seem to have a sufficiently well-defined API that uvloop
can use (or this PR changes said API -- but I suspect there's no API
because none is documented).

I'm not sure I read your message correctly, but my experience with uvloop showed that even tiny deviations from current asyncio behaviour can break other's code and unittests. I think it's good that we now have uvloop with its own tests (and I also run the same tests on asyncio) to stabilize asyncio as a platform.

For instance, this is how I caught a recent regression in `loop.getaddrinfo`.  I had a couple of issues for uvloop where `uvloop.getaddrinfo` didn't support `None` for `port` etc, and asyncio was fine with that. People complained, I fixed that, and today I discovered that it isn't working in latest asyncio too -- something we shouldn't break in a bugfix release.

gvanrossum: My point is that these regressions aren't caught by asyncio's own tests,
and the precise behavior of the APIs that were broken isn't sufficiently
documented in asyncio's documentation. So we're back to developing
user/library code (e.g. uvloop) based on what happens to be implemented
rather than based on a solid specification. This indeed means that almost
any change to asyncio's implementation has the risk of breaking some use of
it...

asyncio/transports.py
```diff
@@ -96,8 +108,12 @@ def writelines(self, list_of_data):
         The default implementation concatenates the arguments and
         calls write() on the result.
         """
-        data = compat.flatten_list_bytes(list_of_data)
-        self.write(data)
+        if not compat.PY34:
```
socketpair: Note, this code is moved from compat.py, since it used only here.

asyncio/compat.py
1st1: To your point -- I'll collect a list of things that I think can be better documented/defined in asyncio documentation. Maybe we can even update the asyncio PEP.

tests/test_selector_events.py
```diff
@@ -718,13 +733,13 @@ def test_close_write_buffer(self):
...
         tr = self.create_transport()
-        tr._buffer.extend(b'1')
+        sock_prefill_buffer(tr, b'1')
         self.loop.add_reader(7, mock.sentinel)
         self.loop.add_writer(7, mock.sentinel)
         tr._force_close(None)
 
         self.assertTrue(tr.is_closing())
-        self.assertEqual(tr._buffer, list_to_buffer())
+        self.assertSequenceEqual([], tr._buffer)
```
socketpair: `assertSequenceEqual` is much better than comparing to `list(tr._buffer)` or `list_to_buffer([...])`! and also I elminate dependency from `deque` in this file!

asyncio/transports.py
socketpair: Finally, I have restore old behaviour...But I'm sure, that calling write on each item should be faster. Will investigate later.

socketpair: About NODELAY. Yes, this is definitely another PR, but since we buffer writes in Python, buffering in kernel have no sense.

AUTHORS
socketpair: fixed

asyncio/selector_events.py
```diff
@@ -704,43 +712,91 @@ def write(self, data):
...
+
+        assert 0 <= n <= self._buffer_size
+
+        while n > 0:
+            chunk = self._buffer.popleft()
+            chunk_len = len(chunk)
+            self._buffer_size -= chunk_len
+            n -= chunk_len
+
+        if n < 0:
```
socketpair: This case is covered by `test_write_ready_partial`

asyncio/unix_events.py
```diff
@@ -513,39 +517,84 @@ def write(self, data):
...
 
-            self._buffer.append(data)  # Try again later.
+        while n > 0:
+            chunk = self._buffer.popleft()
+            chunk_length = len(chunk)
+            self._buffer_size -= chunk_length
+            n -= chunk_length
+
+        if n < 0:
+            # only part of chunk was written, so push unread part of it
```
socketpair: coveredy by `test__write_ready_partial`

socketpair:
@1st1 Please review again, and say me what to fix, since I fixed everything known, as I think, including test. (I have also checked code coverage in PyCharm)

asyncio/compat.py
```diff
@@ -1,17 +1,17 @@
...
 
+try:
+    SC_IOV_MAX = os.sysconf(os.sysconf_names['SC_IOV_MAX'])
+except KeyError:
+    # It is not defined how IOV-related syscalls are limited
+    # when SC_IOV_MAX is missing.
+    # MacOS X, FreeBSD and Linux typically have value of 1024
+    SC_IOV_MAX = 16
 
-def flatten_list_bytes(list_of_data):
```
socketpair: this code moved directly to place where it was used.

socketpair:
@1st1 In another PR I want try to merge pipes code and socket code. Today these codes are just copy-pasted, and this is bother me a lot.

1st1:
> @1st1 Please review again, and say me what to fix, since I fixed everything known, as I think, including test. (I have also checked code coverage in PyCharm)

Please also check that you have no lines longer than 79 characters.  I've seen a few earlier.

> @1st1 In another PR I want try to merge pipes code and socket code. Today these codes are just copy-pasted, and this is bother me a lot.

It would depend on the size of that PR, but maybe we'll do that after `3.5.2`.

socketpair:
Well, I have checked line lengths using this:

`git diff upstr/master HEAD asyncio | egrep '^[+].{80,}'`

1st1:
Should be 79 per PEP 8, no?

socketpair:
@1st1 test show that no lines having 80 symbols :). So, all lines added have 79 symbols or less.

socketpair:
@1st1 https://github.com/ogrisel/python-appveyor-demo/blob/master/appveyor.yml allows to check under windows. Or I'm wrong ?

1st1:
Yep.  Good idea to add it, I'll take a look later.

socketpair:
@1st1 I move ot to separate issue.

asyncio/transports.py
```diff
@@ -87,7 +87,19 @@ def write(self, data):
 
         This does not block; it buffers the data and arranges for it
         to be sent out asynchronously.
+
+        If passed argument is bytearray, it is not allowed to change the  size
+        of that object while write is in progress. Violation of that lead
+        to undefined behaviour.
         """
+
+        # In order to lock size of bytearray, we may wrap bytearray to
```
socketpair: @1st1 @gvanrossum 
This is API change (!) in the past, one may pass mutable buffer and safely modify it. Now it is forbidden. Is it acceptable ?

1st1: My opinion on this: `if type(data) is not bytes: data = memoryview(data)`.

The performance impact is negligible (since `writev` and `sendmsg` use buffer API too), but it will make it less error prone.  Moreover, if your code modifies the bytearray and works, it can break at any time, since it's not guaranteed that your data is sent on the first try in current asyncio.

socketpair: Actually, it degrades performance....about 50-60 MB/sec on my PC.

1st1: 50-60 MB/sec relative to what? And what's benchmark? If it's the one that makes 5000 writes in a row of course it will run a tad slower, but that's a very unrealistic scenario.

socketpair: Even `memoryview` does not protect from this.

```
xxx = bytearray(65536)

somesock.recv_into(xxx)
stream.write(xxx)
somesock.recv_into(xxx)
stream.write(xxx)
```

And so, we should convert to `bytes()`, since Python do not have `.freeze()` interface (http://bugs.python.org/issue27070).

If conversion would be added, performance will be brought down even more.

Why we should support stupid code? Modifying still-not-written buffers was never allowed, so if one rely on specific undocumented implementation, he must suffer.

1st1: > Even memoryview does not protect from this.

1st1: > Even memoryview does not protect from this.

Wow, I was absolutely under impression that it would. Maybe `Py_buffer` C API stuff does... Let me get back to you on this.

> Why we should support idiotic code? Modifying still-not-written buffers was never allowed, so if one rely on specific undocumented implementation, he must suffer.

Probably we shouldn't care. If `memoryview` doesn't work, I'd be OK to not to do anything here.

socketpair: See in action:

```
In [1]: x = bytearray(b'qwe')
In [2]: y = memoryview(x)
In [5]: y[0]=65
In [6]: x[1]=66
In [7]: x
Out[7]: bytearray(b'ABe')
In [9]: bytes(y)
Out[9]: b'ABe'
```

socketpair: @gvanrossum Do you know info about https://www.python.org/dev/peps/pep-0351 ?

http://bugs.python.org/issue27070 says that it was rejected.

This PEP is exactly what I want.

socketpair:
@1st1 any progress on this PR? Should I fix something ?

socketpair:
@1st1 @gvanrossum @methane bump. I want this to be merged :)

1st1:
@socketpair yeah, I'll give it another review soon. Sorry, I've a lot on my plate right now. 

@haypo would you like to take a break from hacking on CPython internals and co-review this patch?  It's pretty big and I'm a bit afraid to push it to 3.5.2 alone.

socketpair:
Only one thing that I afraid with this patch - is ability to totally destroy internal state of the stream object when passed buffers changed externally while still not written by the stream. It was not documented earlier if that is allowed, but it is changed. Only one cause of such behaviour - is performance. Unfortunatelly Python have neither copy-on-write buffers nor API for locking buffers.

socketpair:
Also, someone should port these changes to WIndows-implementation. I will not even try to do that.

asyncio/transports.py
methane: > Even memoryview does not protect from this.
> 
> xxx = bytearray(65536)
> 
> somesock.recv_into(xxx)
> stream.write(xxx)
> somesock.recv_into(xxx)
> stream.write(xxx)
> And so, we should convert to bytes(), since Python do not have .freeze() interface (http://bugs.python.org/issue27070).

Using single bytearray instead of list (or deque) of memoryview for buffer can allow such use cases, too.
Additionally, bytearray can protect `len(memoryview)` pitfall.

```
In [1]: import array
In [2]: a = array.array('h', [1,2,3,4])
In [3]: mv = memoryview(a)
In [4]: len(mv)
Out[4]: 4
In [5]: ba = bytearray()
In [7]: ba += mv
In [8]: len(ba)
Out[8]: 8
In [9]: ba
Out[9]: bytearray(b'\x01\x00\x02\x00\x03\x00\x04\x00')
```

Unless deque[Union[memoryview, bytes, bytearray]] is bit faster, I prefer bytearray for it's safety and simplicity.
How much this pull request is faster than single bytearray approach?

socketpair: According to benchmark tool added in this PR (see first commit):

(editted to show only difference between `bytearray` approach and `deque [bytes]`. This is exactly what happen for sockets)

```
# original
5 buffers by 10.00 MB : 0.25 seconds (199.85 MB/sec)
5120 buffers by 0.01 MB : 0.18 seconds (274.86 MB/sec)
5 buffers by 10.00 MB (with intermediate drains): 0.19 seconds (261.55 MB/sec)
5120 buffers by 0.01 MB (with intermediate drains): 0.12 seconds (421.09 MB/sec)

# patched
5 buffers by 10.00 MB : 0.11 seconds (444.23 MB/sec)     [2.22 times faster]
5120 buffers by 0.01 MB : 0.10 seconds (479.93 MB/sec)   [1.74 times faster]
5 buffers by 10.00 MB (with intermediate drains): 0.11 seconds (453.50 MB/sec) [1.73 times faster]
5120 buffers by 0.01 MB (with intermediate drains): 0.12 seconds (431.83 MB/sec) [1.02 times faster]
```

asyncio/compat.py
```diff
@@ -1,17 +1,17 @@
 """Compatibility helpers for the different Python versions."""
 
+import os
 import sys
 
 PY34 = sys.version_info >= (3, 4)
 PY35 = sys.version_info >= (3, 5)
 
+try:
+    SC_IOV_MAX = os.sysconf(os.sysconf_names['SC_IOV_MAX'])
```
socketpair: Thanks to appveyor checking, Windows build is broken:

https://ci.appveyor.com/project/1st1/asyncio/build/1.0.5/job/m2humsxc1tp1jxya

will fix

socketpair:
What the hell! Python 3.4 for windows have no `sendmsg()` ?!

socketpair:
How to fix tests? behaviour of internal logic is different for Windows and Linux. It would be mess of code. Please advise me.

1st1:
> Python for Windows have no sendmsg() ?

Looks like you have to detect that and provide a fallback code that uses `send`.

socketpair:
YAY! tests passed! But wait. During fixing tests I have added `if sys.platform == 'win32'` and 3.5 python Succeeded, but 3.4 fail! seems Python3.4 HASE sendmsg. What the hell ! Will investigate.

Also I have rebased and fix conflicts.

socketpair:
That was VERY strange bug that I can not understand. And don't want to investigate. Everything works as expected now. Please review and merge.

socketpair:
@1st1 @methane @gvanrossum any progress? Seems everything known is fixed, including Windows issues.

1st1:
I'll give this PR a final review soon.  We've missed this PR for 3.5.2 anyways, so there is no rush to have this committed.

socketpair:
ping

asyncio/selector_events.py
```diff
@@ -711,45 +718,106 @@ def write(self, data):
...
 
     def _write_ready(self):
-        assert self._buffer, 'Data should not be empty'
+        assert self._buffer and self._buffer_size > 0, 'Data must not be empty'
 
         if self._conn_lost:
             return
         try:
-            n = self._sock.send(self._buffer)
+            if compat.HAS_SENDMSG:
```
socketpair: HAS or HAVE ? :)

1st1: HAS is better.

1st1:
> ping

Sorry, Mark, I'm busy with other things.  This is quite an intrusive change and it has to be merged with extra care.  And we have plenty of time before 3.6.

socketpair:
@1st1 I can split PR to small parts. After each part, everything will be consistent. Should I do that ?

1st1:
> @1st1 I can split PR to small parts. After each part, everything will be consistent. Should I do that ?

How do you want to split it?

Actually, I was just looking at this PR again, and I think it's ready.  I need to spend another hour playing with the code, but I'd say it's almost there.

asyncio/selector_events.py
methane: Since caller may change data.  For example:

```python
# Avoid copy as possible
buf = bytearray(4096)
view = memoryview(buf)
while True:
    n = file.readinto(buf)
    if n == 0:
        break
    transport.send(view[:n])
```

methane: And `data = bytes(data)` avoids `len(memoryview)` pitfall, described below too.

https://github.com/python/asyncio/pull/339/files#r64995800

methane:
I think `.sendto()` and `.write()` should `data = bytes(data)` to avoid pitfalls.
If data is bytes already, there is no performance penalty.
If data is bytearray or memoryview, user can reuse buffer to avoid penalty.

asyncio/selector_events.py
socketpair: seems, code above should not be used. It is never guaranteed to work. Since we use internal buferring, we can not eliminate copying in genereal. So, at best, we should steal buffers. i.e. to avoid any copyin, user should write:

```
# Avoid copy as possible
while True:
    view = memoryview(bytearray(4096))
    n = file.readinto(view)
    if n == 0:
        break
    transport.send(view[:n])
```
(did not check that code, but it should work)

methane: >  Since we use internal buferring, we can not eliminate copying in genereal.

I agree.  There should be one copy at least.

> So, at best, we should steal buffers.

I can't agree.  `data = bytes(data)` and `buffer += data` (which buffer is bytearray) can
guard from many errors.
For example, `transport.send(42)` can raise TypeError immediately.
Bug caused by mutating bytearray is very hard to debug.

methane: @1st1 How do you think about this?

I still prefer bytearray approach.  When performance is important, there is uvloop now.

socketpair:
@arthurdarcet can you benchmark streams using included tool (in that PR) on Mac OS X ?

arthurdarcet:
@socketpair sure, do you have some code i could run?

methane:
@socketpair I did it.
https://github.com/python/asyncio/pull/385#issuecomment-234536614

asyncio/selector_events.py
1st1: > @1st1 How do you think about this?

Sorry, I'm away from home (at EuroPython). Will read this discussion once I'm at home, in two-three days.

> I still prefer bytearray approach. When performance is important, there is uvloop now.

We need to find a balance between complexity of code in asyncio & performance. Having uvloop around shouldn't stop us from making asyncio fast too.

socketpair:
I have an idea: add to write functions parameters `gift=True` (or, maybe `gift=False` in order to save compatibility if that was never guaranteed). Inspired by http://man7.org/linux/man-pages/man2/vmsplice.2.html  (SPLICE_F_GIFT)

methane:
I feel it's a premature optimization.

methane:
I've benchmarked on Ubuntu 16.04, with slightly modified `bench_streams.py`.
master branch (bytearray buffer) is faster for smaller (~500 bytes) messages.

master branch:
```
bench_sockets: 10000000 buffers by 10 bytes : 11.32 seconds (8.42 MiB/sec)
bench_sockets: 5000000 buffers by 20 bytes : 5.72 seconds (16.67 MiB/sec)
bench_sockets: 2000000 buffers by 50 bytes : 2.37 seconds (40.22 MiB/sec)
bench_sockets: 1000000 buffers by 100 bytes : 1.26 seconds (75.42 MiB/sec)
bench_sockets: 500000 buffers by 200 bytes : 0.70 seconds (135.37 MiB/sec)
bench_sockets: 200000 buffers by 500 bytes : 0.37 seconds (256.61 MiB/sec)
bench_sockets: 100000 buffers by 1000 bytes : 0.26 seconds (370.33 MiB/sec)
bench_sockets: 50000 buffers by 2000 bytes : 0.20 seconds (474.92 MiB/sec)
bench_sockets: 20000 buffers by 5000 bytes : 0.16 seconds (580.86 MiB/sec)
bench_sockets: 10000 buffers by 10000 bytes : 0.15 seconds (620.15 MiB/sec)
bench_sockets: 5000 buffers by 20000 bytes : 0.15 seconds (631.98 MiB/sec)
bench_sockets: 2000 buffers by 50000 bytes : 0.15 seconds (649.65 MiB/sec)
bench_sockets: 10000000 buffers by 10 bytes (with intermediate drains): 22.57 seconds (4.22 MiB/sec)
bench_sockets: 5000000 buffers by 20 bytes (with intermediate drains): 11.29 seconds (8.45 MiB/sec)
bench_sockets: 2000000 buffers by 50 bytes (with intermediate drains): 4.52 seconds (21.09 MiB/sec)
bench_sockets: 1000000 buffers by 100 bytes (with intermediate drains): 2.31 seconds (41.31 MiB/sec)
bench_sockets: 500000 buffers by 200 bytes (with intermediate drains): 1.19 seconds (80.05 MiB/sec)
bench_sockets: 200000 buffers by 500 bytes (with intermediate drains): 0.52 seconds (183.04 MiB/sec)
bench_sockets: 100000 buffers by 1000 bytes (with intermediate drains): 0.30 seconds (320.99 MiB/sec)
bench_sockets: 50000 buffers by 2000 bytes (with intermediate drains): 0.19 seconds (514.17 MiB/sec)
bench_sockets: 20000 buffers by 5000 bytes (with intermediate drains): 0.12 seconds (814.84 MiB/sec)
bench_sockets: 10000 buffers by 10000 bytes (with intermediate drains): 0.09 seconds (1013.89 MiB/sec)
bench_sockets: 5000 buffers by 20000 bytes (with intermediate drains): 0.08 seconds (1152.17 MiB/sec)
bench_sockets: 2000 buffers by 50000 bytes (with intermediate drains): 0.08 seconds (1257.04 MiB/sec)
```

socketpair:writev branch:
```
bench_sockets: 10000000 buffers by 10 bytes : 15.59 seconds (6.12 MiB/sec)
bench_sockets: 5000000 buffers by 20 bytes : 7.84 seconds (12.16 MiB/sec)
bench_sockets: 2000000 buffers by 50 bytes : 3.17 seconds (30.07 MiB/sec)
bench_sockets: 1000000 buffers by 100 bytes : 1.61 seconds (59.39 MiB/sec)
bench_sockets: 500000 buffers by 200 bytes : 0.85 seconds (111.82 MiB/sec)
bench_sockets: 200000 buffers by 500 bytes : 0.38 seconds (247.81 MiB/sec)
bench_sockets: 100000 buffers by 1000 bytes : 0.23 seconds (419.86 MiB/sec)
bench_sockets: 50000 buffers by 2000 bytes : 0.15 seconds (651.97 MiB/sec)
bench_sockets: 20000 buffers by 5000 bytes : 0.10 seconds (946.71 MiB/sec)
bench_sockets: 10000 buffers by 10000 bytes : 0.08 seconds (1131.09 MiB/sec)
bench_sockets: 5000 buffers by 20000 bytes : 0.08 seconds (1236.96 MiB/sec)
bench_sockets: 2000 buffers by 50000 bytes : 0.07 seconds (1324.88 MiB/sec)
bench_sockets: 10000000 buffers by 10 bytes (with intermediate drains): 26.38 seconds (3.61 MiB/sec)
bench_sockets: 5000000 buffers by 20 bytes (with intermediate drains): 12.92 seconds (7.38 MiB/sec)
bench_sockets: 2000000 buffers by 50 bytes (with intermediate drains): 5.34 seconds (17.87 MiB/sec)
bench_sockets: 1000000 buffers by 100 bytes (with intermediate drains): 2.38 seconds (40.15 MiB/sec)
bench_sockets: 500000 buffers by 200 bytes (with intermediate drains): 1.24 seconds (77.00 MiB/sec)
bench_sockets: 200000 buffers by 500 bytes (with intermediate drains): 0.55 seconds (174.32 MiB/sec)
bench_sockets: 100000 buffers by 1000 bytes (with intermediate drains): 0.31 seconds (311.23 MiB/sec)
bench_sockets: 50000 buffers by 2000 bytes (with intermediate drains): 0.19 seconds (498.30 MiB/sec)
bench_sockets: 20000 buffers by 5000 bytes (with intermediate drains): 0.12 seconds (804.88 MiB/sec)
bench_sockets: 10000 buffers by 10000 bytes (with intermediate drains): 0.09 seconds (1005.35 MiB/sec)
bench_sockets: 5000 buffers by 20000 bytes (with intermediate drains): 0.08 seconds (1141.88 MiB/sec)
bench_sockets: 2000 buffers by 50000 bytes (with intermediate drains): 0.08 seconds (1258.54 MiB/sec)
```

socketpair:
1. And, seems, I know why. Python build `iov` array every time :(
2. Seems, we should test on real use cases. Actually, it will be not such big buffers with count of 10 :

`writev`:
pros:
* true zero copy
* slightly faster
* use syscall that was intended exactly for that
cons:
* writev is not supported in Windows
* adds restrictions on passing of mutable buffers
* complex and fragile code

`bytearray`
pros:
* much simplier
* do not require hack for Windows
cons:
* slightly slower

@gvanrossum @1st1 @methane 
I can not choose by myself. :( please help me to make decision :)

methane:
Another cons of `writev`: `TypeError` may happen in `_write_ready()`, not `write()`.

Checking "bytes-like" object without copy is very hard in pure Python.
I think following code is same to `PyObject_GetBuffer(data, &view, PyBUF_SIMPLE)`,
but I'm not sure.

```py
mv = memoryview(data)
if mv.itemsize != 1 or not mv.contiguous:
    raise TypeError
```

`bytearray += data` raises TypeError when data is not bytes-like object.

methane:
My preference is, use bytearray() in pure Python.  But make it replacable.

Someone can implement more smart buffer like below in C.

```py
def __init__(self):
    self.queue = deque()

def __iadd__(self, data):
    if type(data) is bytes and len(data) > 200:
        self.queue.append(data)
        return
    if not self.queue or type(self.queue[-1]) is bytes:
        self.queue.append(bytearray())
    self.queue[-1] += data
```

socketpair:
Copy-pasting here:

small data:

```
bytearray:
bench_sockets: 10000000 buffers by 10 bytes : 11.32 seconds (8.42 MiB/sec)
bench_sockets: 1000000 buffers by 100 bytes : 1.26 seconds (75.42 MiB/sec)

writev:
bench_sockets: 10000000 buffers by 10 bytes : 15.59 seconds (6.12 MiB/sec)
bench_sockets: 1000000 buffers by 100 bytes : 1.61 seconds (59.39 MiB/sec)
```

large data:

```
bytearray:
bench_sockets: 100000 buffers by 1000 bytes : 0.26 seconds (370.33 MiB/sec)
bench_sockets: 10000 buffers by 10000 bytes : 0.15 seconds (620.15 MiB/sec)
bench_sockets: 2000 buffers by 50000 bytes : 0.15 seconds (649.65 MiB/sec)

writev:
bench_sockets: 100000 buffers by 1000 bytes : 0.23 seconds (419.86 MiB/sec)
bench_sockets: 10000 buffers by 10000 bytes : 0.08 seconds (1131.09 MiB/sec)
bench_sockets: 2000 buffers by 50000 bytes : 0.07 seconds (1324.88 MiB/sec)
```

socketpair:
@gvanrossum @1st1 What do you think about all that ?

socketpair:
I understand, that simplicity of code is important thing. In real use cases (i.e. small amount of data and small count of chunks) both implementations give the same result.

1st1:
Mark, I'm closing this PR. I think we'd want to have vectorized writes in 3.7, so when you have a new patch please open a new PR.

PR #344, opened 2016-05-07
ease the cert failed regex

pjenvey:
These are too specific to CPython which emits:

"[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:600)"


1st1:
[Merged](https://hg.python.org/cpython/rev/deb8e076a096) in CPython.

Issue #340, opened 2016-05-04
Add get_exeption_handler()


sametmax:
We have set_exeption_handler(), but no getter. Currently we use _exception_handler to obtain the previous one, but it's already causing [some troubles](https://github.com/MagicStack/uvloop/issues/3#issuecomment-216704907). 

A property would actually be even more idiomatic than a get/set method.

1st1:
Congrats, we now have the `get_exception_handler` method. I'll add it to uvloop tomorrow.

PR #345, opened 2016-05-14
Add 'loop.get_exception_handler()' method.

1st1:
This PR implements a new event loop method `get_exception_handler()`. All other `set_*` methods have a corresponding `get_*` method. See also issue #340.


1st1:
Thank you, Guido! I'll update CPython docs/whatsnew tomorrow.

PR #346, opened 2016-05-16
locks: Fix inconsistency cancelling Condition.wait.

dcoles:
Issue #22970: Cancelling wait() after notification leaves Condition in an
inconsistent state.

This can occur if the waiting task has been notified but has not reacquired the
associated lock. Instead of cancelling waiting for the notification, the cancel
call will cancel the lock reacquisition, returning control to the calling task
without the lock being held (a violation of Condition.wait contract).

### DIFF
--- a/tests/test_locks.py
+++ b/tests/test_locks.py
@@ -457,6 +457,31 @@ class ConditionTests(test_utils.TestCase):
         self.assertFalse(cond._waiters)
         self.assertTrue(cond.locked())
 
+    def test_wait_cancel_contested(self):
+        cond = asyncio.Condition(loop=self.loop)
+
+        self.loop.run_until_complete(cond.acquire())
+        self.assertTrue(cond.locked())
+
+        wait_task = asyncio.Task(cond.wait(), loop=self.loop)
+        test_utils.run_briefly(self.loop)
+        self.assertFalse(cond.locked())
+
+        # Notify, but contest the lock before cancelling
+        self.loop.run_until_complete(cond.acquire())
+        self.assertTrue(cond.locked())
+        cond.notify()
+        self.loop.call_soon(wait_task.cancel)
+        self.loop.call_soon(cond.release)
+
+        try:
+            self.loop.run_until_complete(wait_task)
+        except asyncio.CancelledError:
+            # Should not happen, since no cancellation points
+            pass
+
+        self.assertTrue(cond.locked())
+
     def test_wait_unacquired(self):
         cond = asyncio.Condition(loop=self.loop)
         self.assertRaises(
--- a/asyncio/locks.py
+++ b/asyncio/locks.py
@@ -329,7 +329,13 @@ class Condition(_ContextManagerMixin):
                 self._waiters.remove(fut)
 
         finally:
-            yield from self.acquire()
+            # Must reacquire lock even if wait is cancelled
+            while True:
+                try:
+                    yield from self.acquire()
+                    break
+                except futures.CancelledError:
+                    pass
 
     @coroutine
     def wait_for(self, predicate):


tests/test_locks.py
```diff
@@ -457,6 +457,31 @@ def test_wait_cancel(self):
...
+        # Notify, but contest the lock before cancelling
+        self.loop.run_until_complete(cond.acquire())
+        self.assertTrue(cond.locked())
+        cond.notify()
+        self.loop.call_soon(wait_task.cancel)
+        self.loop.call_soon(cond.release)
+
+        try:
+            self.loop.run_until_complete(wait_task)
+        except asyncio.CancelledError:
```
dcoles: Originally I had a `self.assertRaises(asyncio.CancelledError, ...)` here, but since we successfully notified this task before cancellation no exception should be raises. However if the fix is not applied then a CancelledError will be raised so we have to be prepared to handle it so we can check the lock state afterwards.

tests/test_locks.py
```diff
@@ -457,6 +457,31 @@ def test_wait_cancel(self):
...
+            # Should not happen, since no cancellation points
+            pass
```
1st1: But why do we have the `try..except` here (if it shouldn't happen)? Maybe instead of `pass` you want `self.fail('explanation')`?

1st1:
@dcoles I've merged this PR in 27218fa7cc5614c4f08a34712116ddc5489ec1f6.  It should make it to CPython 3.5.2.  Thanks!

Although my question about the unit test is still open.

tests/test_locks.py
dcoles: The reason this was just a pass rather than a hard-fail is so that we reach the `assertTrue(cond.locked())` to ensure that this patch resolves the issue (failing before the patch is applied and then not afterwards). With a fail in the exception handler, we wouldn't be showing the existence of issue 22970.

1st1: Should that try-except be replaced with self.assertRaises?

dcoles:
Not a problem. Thanks taking the time in reviewing and looking into the issue.

In the future when you merge a contribution, could you please preserve the authorship information by either directly merging this branch, cherry picking or using --author flag when committing. Besides preserving contribution information, my employer requires the use of my work email address when contributing to open source projects.

1st1:
> In the future when you merge a contribution, could you please preserve the authorship information by either directly merging this branch, cherry picking or using --author flag when committing. Besides preserving contribution information, my employer requires the use of my work email address when contributing to open source projects.

Next time please indicate that as early as possible.  Your patch was merged in the same way how 99% patches usually get merged to CPython and asyncio.

gvanrossum:
Hm, for the asyncio repo here don't we usually use the "Merge pull request" button on GitHub? That would have done what David asked for IIUC. The PR is still open so apparently it didn't happen that way.

1st1:
I usually merge by hand -- fetching the repo, rebasing the commit to avoid merges, and then pushing it (authorship is preserved that way).

This time I simply downloaded this pr as a patch and committed it to cpython and asyncio attributing David in the commit message... I can revert my commit and push again FWIW.

gvanrossum:
GitHub has a relatively new "squash merge" feature in the UI that avoids
the annoying merges. I rarely merge by hand any more (only when there's
something I want to clean up and the author is not responsive).

1st1:
I think I've enabled it for asyncio repo. Ok, I'll try that button next time :)

tests/test_locks.py
dcoles: Unfortunately you then get the opposite problem where no exception is raised once bug is fixed. Stuck between these two unfortunate alternatives, I decided that using `pass` was cleanest.

dcoles:
Since you asked me to create a PR here on github, I was under the assumption that that's the way it would have been merged. Since my employer claims copyright on pretty much any code I write (but are fine with contributions to open source software) they'd like to see contributions from my dcoles@google.com address. There's also at least one other prominent 'David Coles' contributing to open source.

(Plus it gives me a bit of a warm fuzzy feeling to see git/github tell me I've contributed to Python/Asyncio :smile: ).

I think the reason for the missing authorship is, as you said, just directly applying the patch (to avoid the merge commit). No harm done. I'm mostly happy to see that this will be fixed in 3.5.2!

1st1:
Sorry again, David, and thanks a lot for your contribution. I think that once we migrate CPython repo to github, it will be an established practice to merge commits preserving the authorship.

Issue #353, opened 2016-05-23
getaddrinfo doesn't accept port service names


1st1:
`socket.getaddrinfo` and `loop.getaddrinfo` in Python 3.5.1 can accept service names for its `port` argument, i.e.

```python
>>> socket.getaddrinfo('127.0.0.1', 'http')
[(<AddressFamily.AF_INET: 2>, <SocketKind.SOCK_DGRAM: 2>, 17, '', ('127.0.0.1', 80)),
 (<AddressFamily.AF_INET: 2>, <SocketKind.SOCK_STREAM: 1>, 6, '', ('127.0.0.1', 80))]
```

In 3.5.2 with our custom `_ipaddr_info` such calls will fail with an exception:

```python
>>> import asyncio, socket
>>> loop = asyncio.get_event_loop()
>>> loop.run_until_complete(loop.getaddrinfo('127.0.0.1', 'http', type=socket.SOCK_STREAM))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/yury/dev/py/asyncio/asyncio/base_events.py", line 589, in getaddrinfo
    info = _ipaddr_info(host, port, family, type, proto)
  File "/Users/yury/dev/py/asyncio/asyncio/base_events.py", line 108, in _ipaddr_info
    port = int(port)
ValueError: invalid literal for int() with base 10: 'http'
```

/cc @ajdavis

gvanrossum:
Honestly I thought that such service names were a thing of the past. But ah well...

1st1:
Looks like we need to use https://docs.python.org/3/library/socket.html#socket.getservbyname if we can't cast port to an int.

gvanrossum:
You mean a numeric string like '80' would also be accepted?

1st1:
Yes, and event `b'80'`, since `socket.getaddrinfo` accepts it. Apparently people use that.

ajdavis:
This little project still has more corners of socket programming to teach me! I've dealt with '80' and b'80' already, this last patch also handles 'http' and b'http'.

1st1:
Closing this issue -- the relevant PR has been merged.

PR #354, opened 2016-05-23
loop.getaddrinfo accepts service names like "http" for port.

ajdavis:
loop.getaddrinfo accepts service names like "http" for port.

### DIFF
--- a/asyncio/base_events.py
+++ b/asyncio/base_events.py
@@ -102,20 +102,26 @@ def _ipaddr_info(host, port, family, type, proto):
     else:
         return None
 
-    if port in {None, '', b''}:
+    if port is None:
         port = 0
     elif isinstance(port, bytes):
-        try:
-            port = int(port)
-        except ValueError:
-            # Might be a service name like b"http".
-            port = socket.getservbyname(port.decode('ascii'))
+        if port == b'':
+            port = 0
+        else:
+            try:
+                port = int(port)
+            except ValueError:
+                # Might be a service name like b"http".
+                port = socket.getservbyname(port.decode('ascii'))
     elif isinstance(port, str):
-        try:
-            port = int(port)
-        except ValueError:
-            # Might be a service name like "http".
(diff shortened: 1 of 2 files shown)


ajdavis:
Fixes #353 .

asyncio/base_events.py
```diff
@@ -104,8 +104,14 @@ def _ipaddr_info(host, port, family, type, proto):
...
         port = 0
-    elif isinstance(port, (bytes, str)):
+    elif isinstance(port, bytes):
         port = int(port)
+    elif isinstance(port, str):
+        try:
+            port = int(port)
+        except ValueError:
+            # Might be a service name like "http".
+            port = socket.getservbyname(port)
```
1st1: This works only when `port` is an `str`, right?  How does the `port=b'http'` test pass?

ajdavis: Good catch! Fixed.

1st1: Ah, I see. But why don't we just convert bytes to str?

`socket.getaddrinfo('127.0.0.1', b'http')` works so should our code.

asyncio/base_events.py
```diff
@@ -104,8 +104,18 @@ def _ipaddr_info(host, port, family, type, proto):
 
     if port in {None, '', b''}:
```
1st1: While you're here, could you please fix a `BytesWarning`:

Please replace `if port in {None, '', b''}:` with `if port is None:`; 
add `if port == '': port = 0 else:` to the `elif isinstance(port, str)`, 
do the same for bytes.

ajdavis: Done.

1st1:
Merged by hand in 2959a97823cdf2905e343efc7f3ec8dc013970f7

Thanks a lot, Jesse.

1st1:
See also https://hg.python.org/cpython/rev/3ec208c01418

PR #355, opened 2016-05-25
Ensure get_event_loop returns the running loop when called in a coroutine

vxgmichel:
This PR follows up on the [issue 26969](http://bugs.python.org/issue26969).

It is meant to ensure that `asyncio.get_event_loop` returns the running loop when called in a coroutine (or loop callbacks). This could simplify all the coroutines that includes a `loop` optional argument by removing it.

This is done by modifying  the interface of `AbstractEventLoopPolicy`:
- `get/set_event_loop` is renamed to `get/set_default_loop`
- `get/set_running_loop` is added
- `get_event_loop` now uses the running loop if available, and the default loop otherwise

`BaseEventLoopPolicy` is updated, and adds a few features:
- it does not allow to set a running loop if another one is already set
- it issues warnings if the running loop doesn't correspond to the default loop

A context manager is also added to `AbstractEventLoopPolicy` and `AbstractEventLoop` to set and clear the running loop in a safe manner. This might help for other event loop implementations.

### DIFF
--- a/asyncio/base_events.py
+++ b/asyncio/base_events.py
@@ -342,11 +342,15 @@ class BaseEventLoop(events.AbstractEventLoop):
         self._set_coroutine_wrapper(self._debug)
         self._thread_id = threading.get_ident()
         try:
-            with self._running_context():
+            policy = events.get_event_loop_policy()
+            policy.set_running_loop(self)
+            try:
                 while True:
                     self._run_once()
                     if self._stopping:
                         break
+            finally:
+                policy.set_running_loop(None)
         finally:
             self._stopping = False
             self._thread_id = None
--- a/asyncio/base_events.py
+++ b/asyncio/base_events.py
@@ -342,10 +342,11 @@ class BaseEventLoop(events.AbstractEventLoop):
         self._set_coroutine_wrapper(self._debug)
         self._thread_id = threading.get_ident()
         try:
-            while True:
-                self._run_once()
-                if self._stopping:
-                    break
+            with self._running_context():
+                while True:
+                    self._run_once()
+                    if self._stopping:
+                        break
         finally:
             self._stopping = False
             self._thread_id = None
--- a/asyncio/base_events.py
+++ b/asyncio/base_events.py
@@ -339,22 +339,20 @@ class BaseEventLoop(events.AbstractEventLoop):
         self._check_closed()
         if self.is_running():
             raise RuntimeError('Event loop is running.')
+        policy = events.get_event_loop_policy()
+        policy.set_running_loop(self)
         self._set_coroutine_wrapper(self._debug)
         self._thread_id = threading.get_ident()
         try:
-            policy = events.get_event_loop_policy()
-            policy.set_running_loop(self)
-            try:
-                while True:
-                    self._run_once()
-                    if self._stopping:
-                        break
-            finally:
-                policy.set_running_loop(None)
+            while True:
+                self._run_once()
+                if self._stopping:
+                    break
         finally:
             self._stopping = False
             self._thread_id = None
             self._set_coroutine_wrapper(False)
+            policy.set_running_loop(None)
 
     def run_until_complete(self, future):
         """Run until the Future is done.
--- a/asyncio/base_events.py
+++ b/asyncio/base_events.py
@@ -340,10 +340,10 @@ class BaseEventLoop(events.AbstractEventLoop):
         if self.is_running():
             raise RuntimeError('Event loop is running.')
         policy = events.get_event_loop_policy()
-        policy.set_running_loop(self)
         self._set_coroutine_wrapper(self._debug)
         self._thread_id = threading.get_ident()
         try:
+            policy.set_running_loop(self)
             while True:
                 self._run_once()
                 if self._stopping:
@@ -351,8 +351,8 @@ class BaseEventLoop(events.AbstractEventLoop):
         finally:
             self._stopping = False
             self._thread_id = None
-            self._set_coroutine_wrapper(False)
             policy.set_running_loop(None)
+            self._set_coroutine_wrapper(False)
 
     def run_until_complete(self, future):
         """Run until the Future is done.
--- a/asyncio/events.py
+++ b/asyncio/events.py
@@ -16,6 +16,8 @@ import subprocess
 import sys
 import threading
 import traceback
+import warnings
+import contextlib
 
 from asyncio import compat
 
@@ -507,28 +509,52 @@ class AbstractEventLoop:
     def set_debug(self, enabled):
         raise NotImplementedError
 
+    # Running context
+
+    def _running_context(self):
+        return get_event_loop_policy().running_loop_context(self)
+
 
 class AbstractEventLoopPolicy:
     """Abstract policy for accessing the event loop."""
 
-    def get_event_loop(self):
-        """Get the event loop for the current context.
+    def get_default_loop(self):
+        """Get the default event loop for the current context.
 
         Returns an event loop object implementing the BaseEventLoop interface,
         or raises an exception in case no event loop has been set for the
         current context and the current policy does not specify to create one.
 
-        It should never return None."""
+        It should never return None.
+        """
         raise NotImplementedError
 
-    def set_event_loop(self, loop):
-        """Set the event loop for the current context to loop."""
+    def set_default_loop(self, loop):
+        """Set the default event loop for the current context to loop."""
+        raise NotImplementedError
+
+    def get_running_loop(self):
+        """Get the running event loop running for the current context, if any.
+
+        Returns an event loop object implementing the BaseEventLoop interface.
+        If no running loop is set, it returns None.
+        """
+        raise NotImplementedError
+
+    def set_running_loop(self, loop):
+        """Set the running event loop for the current context.
+
+        The loop argument can be None to clear the former running loop.
+        This method should be called by the event loop itself to set the
+        running loop when it starts, and clear it when it's done.
+        """
         raise NotImplementedError
 
     def new_event_loop(self):
         """Create and return a new event loop object according to this
-        policy's rules. If there's need to set this loop as the event loop for
-        the current context, set_event_loop must be called explicitly."""
+        policy's rules. If there's need to set this loop as the default loop
+        for the current context, set_event_loop must be called explicitly.
+        """
         raise NotImplementedError
 
     # Child processes handling (Unix only).
@@ -540,6 +566,37 @@ class AbstractEventLoopPolicy:
     def set_child_watcher(self, watcher):
         """Set the watcher for child processes."""
         raise NotImplementedError
+
+    # Non-abstract methods
+
+    def get_event_loop(self):
+        """Return the running event loop if any, and the default event
+        loop otherwise.
+        """
+        running_loop = self.get_running_loop()
+        if running_loop is not None:
+            return running_loop
+        return self.get_default_loop()
+
+    def set_event_loop(self, loop):
+        """Set the default event loop if the former loop is not currently
+        running.
+        """
+        if self.get_running_loop() is not None:
+            raise RuntimeError('The former loop is currently running')
+        self.set_default_loop(loop)
+
+    @contextlib.contextmanager
+    def running_loop_context(self, loop):
+        """Convenience context to set and clear the given loop as running
+        loop. It is meant to be used inside the loop run methods.
+        """
+        assert isinstance(loop, AbstractEventLoop)
+        self.set_running_loop(loop)
+        try:
+            yield
+        finally:
+            self.set_running_loop(None)
 
 
 class BaseDefaultEventLoopPolicy(AbstractEventLoopPolicy):
@@ -556,16 +613,23 @@ class BaseDefaultEventLoopPolicy(AbstractEventLoopPolicy):
     """
 
     _loop_factory = None
+    _warnings = True
 
     class _Local(threading.local):
         _loop = None
+        _running_loop = None
         _set_called = False
 
     def __init__(self):
         self._local = self._Local()
 
-    def get_event_loop(self):
-        """Get the event loop.
+    def warn(self, *args):
+        if self._warnings:
+            raise RuntimeError(*args)
+            warnings.warn(*args)
+
+    def get_default_loop(self):
+        """Get the default event loop for the current thread.
 
         This may be None or an instance of EventLoop.
         """
@@ -578,11 +642,37 @@ class BaseDefaultEventLoopPolicy(AbstractEventLoopPolicy):
                                % threading.current_thread().name)
         return self._local._loop
 
-    def set_event_loop(self, loop):
-        """Set the event loop."""
+    def set_default_loop(self, loop):
+        """Set the default event loop for the current thread."""
         self._local._set_called = True
         assert loop is None or isinstance(loop, AbstractEventLoop)
         self._local._loop = loop
+
+    def get_running_loop(self):
+        """Get the running event loop for the current thread if any.
+
+        This may be None or an instance of EventLoop.
+        """
+        return self._local._running_loop
+
+    def set_running_loop(self, loop):
+        """Set the running event loop for the current thread."""
+        assert loop is None or isinstance(loop, AbstractEventLoop)
+        default_loop = self._local._loop
+        running_loop = self._local._running_loop
+        if running_loop is not None and loop is not None:
+            raise RuntimeError('A loop is already running')
+        # Warnings
+        if loop is not None:
+            if default_loop is None:
+                self.warn(
+                    'Running a loop with no default loop set',
+                    RuntimeWarning)
+            elif loop != default_loop:
+                self.warn(
+                    'Running a loop different from the default loop',
+                    RuntimeWarning)
+        self._local._running_loop = loop
 
     def new_event_loop(self):
         """Create a new event loop.
--- a/asyncio/events.py
+++ b/asyncio/events.py
@@ -547,9 +547,11 @@ class AbstractEventLoopPolicy:
         raise NotImplementedError
 
     def new_event_loop(self):
-        """Create and return a new event loop object according to this
-        policy's rules. If there's need to set this loop as the event loop for
-        the current context, set_event_loop must be called explicitly.
+        """Create and return a new event loop object.
+
+        The loop is created according to the policy's rules.
+        If there is need to set this loop as the event loop for the
+        current context, set_event_loop must be called explicitly.
         """
         raise NotImplementedError
 
--- a/asyncio/events.py
+++ b/asyncio/events.py
@@ -607,7 +607,7 @@ class BaseDefaultEventLoopPolicy(AbstractEventLoopPolicy):
         return self._local._loop
 
     def set_event_loop(self, loop):
-        """Set the default event loop for the current thread."""
+        """Set the event loop for the current thread."""
         self._local._set_called = True
         assert loop is None or isinstance(loop, AbstractEventLoop)
         self._local._loop = loop
--- a/asyncio/events.py
+++ b/asyncio/events.py
@@ -511,19 +511,22 @@ class AbstractEventLoop:
 class AbstractEventLoopPolicy:
     """Abstract policy for accessing the event loop."""
 
-    def get_default_loop(self):
-        """Get the default event loop for the current context.
+    def get_event_loop(self):
+        """Get the event loop for the current context.
 
-        Returns an event loop object implementing the BaseEventLoop interface,
-        or raises an exception in case no event loop has been set for the
+        Returns an event loop object implementing the BaseEventLoop interface:
+        - the running loop if it has been set (using set_running_loop)
+        - the loop for the current context otherwise.
+
+        It may also raise an exception in case no event loop has been set for the
         current context and the current policy does not specify to create one.
 
         It should never return None.
         """
         raise NotImplementedError
 
-    def set_default_loop(self, loop):
-        """Set the default event loop for the current context."""
+    def set_event_loop(self, loop):
+        """Set the event loop for the current context."""
         raise NotImplementedError
 
     def get_running_loop(self):
@@ -545,8 +548,8 @@ class AbstractEventLoopPolicy:
 
     def new_event_loop(self):
         """Create and return a new event loop object according to this
-        policy's rules. If there's need to set this loop as the default loop
-        for the current context, set_event_loop must be called explicitly.
+        policy's rules. If there's need to set this loop as the event loop for
+        the current context, set_event_loop must be called explicitly.
         """
         raise NotImplementedError
 
@@ -559,19 +562,6 @@ class AbstractEventLoopPolicy:
     def set_child_watcher(self, watcher):
         """Set the watcher for child processes."""
         raise NotImplementedError
-
-    # Non-abstract methods
-
-    def get_event_loop(self):
-        """Return the running loop if any, and the default loop otherwise."""
-        running_loop = self.get_running_loop()
-        if running_loop is not None:
-            return running_loop
-        return self.get_default_loop()
-
-    def set_event_loop(self, loop):
-        """Set the default event loop for the current context."""
-        self.set_default_loop(loop)
 
 
 class BaseDefaultEventLoopPolicy(AbstractEventLoopPolicy):
@@ -597,11 +587,16 @@ class BaseDefaultEventLoopPolicy(AbstractEventLoopPolicy):
     def __init__(self):
         self._local = self._Local()
 
-    def get_default_loop(self):
-        """Get the default event loop for the current thread.
+    def get_event_loop(self):
+        """Get the event loop for the current context.
 
-        This may be None or an instance of EventLoop.
+        Returns an event loop object implementing the BaseEventLoop interface:
+        - the running loop if it has been set (using set_running_loop)
+        - the loop for the current thread otherwise.
         """
+        running_loop = self.get_running_loop()
+        if running_loop is not None:
+            return running_loop
         if (self._local._loop is None and
             not self._local._set_called and
             isinstance(threading.current_thread(), threading._MainThread)):
@@ -611,7 +606,7 @@ class BaseDefaultEventLoopPolicy(AbstractEventLoopPolicy):
                                % threading.current_thread().name)
         return self._local._loop
 
-    def set_default_loop(self, loop):
+    def set_event_loop(self, loop):
         """Set the default event loop for the current thread."""
         self._local._set_called = True
         assert loop is None or isinstance(loop, AbstractEventLoop)
--- a/asyncio/events.py
+++ b/asyncio/events.py
@@ -523,7 +523,7 @@ class AbstractEventLoopPolicy:
         raise NotImplementedError
 
     def set_default_loop(self, loop):
-        """Set the default event loop for the current context to loop."""
+        """Set the default event loop for the current context."""
         raise NotImplementedError
 
     def get_running_loop(self):
@@ -563,20 +563,14 @@ class AbstractEventLoopPolicy:
     # Non-abstract methods
 
     def get_event_loop(self):
-        """Return the running event loop if any, and the default event
-        loop otherwise.
-        """
+        """Return the running loop if any, and the default loop otherwise."""
         running_loop = self.get_running_loop()
         if running_loop is not None:
             return running_loop
         return self.get_default_loop()
 
     def set_event_loop(self, loop):
-        """Set the default event loop if the former loop is not currently
-        running.
-        """
-        if self.get_running_loop() is not None:
-            raise RuntimeError('The former loop is currently running')
+        """Set the default event loop for the current context."""
         self.set_default_loop(loop)
 
 
--- a/asyncio/events.py
+++ b/asyncio/events.py
@@ -16,8 +16,6 @@ import subprocess
 import sys
 import threading
 import traceback
-import warnings
-import contextlib
 
 from asyncio import compat
 
@@ -509,11 +507,6 @@ class AbstractEventLoop:
     def set_debug(self, enabled):
         raise NotImplementedError
 
-    # Running context
-
-    def _running_context(self):
-        return get_event_loop_policy().running_loop_context(self)
-
 
 class AbstractEventLoopPolicy:
     """Abstract policy for accessing the event loop."""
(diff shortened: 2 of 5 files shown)


asyncio/events.py
```diff
@@ -507,28 +509,52 @@ def get_debug(self):
     def set_debug(self, enabled):
         raise NotImplementedError
 
+    # Running context
+
+    def _running_context(self):
```
1st1: Don't think we need this as a separate method, this is something that users shouldn't ever call directly (even though it's a "private" name they will).

asyncio/events.py
```diff
@@ -578,12 +642,38 @@ def get_event_loop(self):
...
+        """Set the running event loop for the current thread."""
+        assert loop is None or isinstance(loop, AbstractEventLoop)
+        default_loop = self._local._loop
+        running_loop = self._local._running_loop
+        if running_loop is not None and loop is not None:
+            raise RuntimeError('A loop is already running')
+        # Warnings
+        if loop is not None:
+            if default_loop is None:
+                self.warn(
```
1st1: Should this be a warning? Might be an annoying thing in unittests.

asyncio/events.py
```diff
@@ -578,12 +642,38 @@ def get_event_loop(self):
...
+                    'Running a loop with no default loop set',
+                    RuntimeWarning)
+            elif loop != default_loop:
+                self.warn(
```
1st1: And why is this a problem?

asyncio/events.py
```diff
@@ -578,12 +642,38 @@ def get_event_loop(self):
...
         self._local._set_called = True
         assert loop is None or isinstance(loop, AbstractEventLoop)
         self._local._loop = loop
 
+    def get_running_loop(self):
+        """Get the running event loop for the current thread if any.
+
+        This may be None or an instance of EventLoop.
+        """
+        return self._local._running_loop
```
1st1: Do we want to return `None` or raise an exception here?

asyncio/events.py
```diff
@@ -541,6 +567,37 @@ def set_child_watcher(self, watcher):
...
+    def set_event_loop(self, loop):
+        """Set the default event loop if the former loop is not currently
+        running.
+        """
+        if self.get_running_loop() is not None:
+            raise RuntimeError('The former loop is currently running')
+        self.set_default_loop(loop)
+
+    @contextlib.contextmanager
+    def running_loop_context(self, loop):
```
1st1: I don't think we need this. `get_running_loop` and `set_running_loop` is enough.

asyncio/test_utils.py
```diff
@@ -407,6 +407,8 @@ def set_event_loop(self, loop, *, cleanup=True):
         assert loop is not None
         # ensure that the event loop is passed explicitly in asyncio
         events.set_event_loop(None)
+        # Disable event loop policy warnings
+        events.get_event_loop_policy()._warnings = False
```
1st1: So this is how you fix the warnings in tests. The problem is that this is certainly acceptable for asyncio, but not for third party code.  Since I'm not sure I understand the utility of these warnings, I'd just remove them.

asyncio/events.py
vxgmichel: Right, it is probably safer!

asyncio/events.py
vxgmichel: My idea here was to keep `BaseEventLoop.run_forever` simple, and make it easier for other event loop implementations. I can remove it if you think it's not worth it. This is how `run_forever` looks like without the context manager:
```python
    def run_forever(self):
        """Run until stop() is called."""
        self._check_closed()
        if self.is_running():
            raise RuntimeError('Event loop is running.')
        self._set_coroutine_wrapper(self._debug)
        self._thread_id = threading.get_ident()
        try:
            policy = get_event_loop_policy()
            policy.set_running_loop(self)
            try:
                while True:
                    self._run_once()
                    if self._stopping:
                        break
            finally:
                policy.set_running_loop(None)
        finally:
            self._stopping = False
            self._thread_id = None
            self._set_coroutine_wrapper(False)
```

asyncio/events.py
vxgmichel: I think `None` is better for different reasons. The most obvious is this one:
```
policy.set_running_loop(None)             # Clear the running loop
assert policy.get_running_loop() == None  # This test is expected to pass
```

asyncio/events.py
vxgmichel: My idea here is to take profit of the policy mecanism to make the default policy a bit stricter than required, in order to encourage users to keep a one-to-one mapping between threads and loops. For instance:

```python
# OK
loop.run_until_complete(coro())
# OK
loop = CustomEventLoop()
asyncio.set_event_loop(loop)
loop.run_until_complete(coro())
# Warning
loop = asyncio.new_event_loop()
loop.run_until_complete(coro())
```

Maybe a warning is not the best solution, or maybe it is not worth it at all. 
Do you know any use case for which using a loop without setting it makes sense?

asyncio/events.py
1st1: If you re-use the outer try-except, it's not going to be that ugly, and it's still easy to use for alternative implementations (can't see any problems to add this to uvloop, for instance). I'd really prefer to keep the number of new APIs to the minimum, so please remove the `running_loop_context` manager.

asyncio/events.py
1st1: Agree, let's keep the current behaviour (returning `None`).

asyncio/events.py
1st1: > Maybe warnings are not the best solution, or maybe it is not worth it at all. 

I'd say this is worth thinking about, but not for 3.5.2. Let's make this PR as simple as possible.

> Do you know any use case for which using a loop without setting it makes sense?

unittests mainly, where people frequently call `set_event_loop(None)` in `unittest.setUp` to make sure that the loop is passed explicitly everywhere in their library.  As for the real code - I don't know, I've never used more than one loop in one process.

1st1:
@vxgmichel Thanks for working on this! There is a chance we can merge this before 3.5.2. Please update the PR.

asyncio/events.py
vxgmichel: All right I'll remove it. However I don't think it is possible to re-use the outer try-except since `policy.set_running_loop` might raise an exception and prevent the run to be cleaned up correctly.

vxgmichel:
@1st1 I just removed the context manager and the warnings. Should I also revert the [simplification for SleepTests and TimeoutTests](https://github.com/python/asyncio/pull/355/commits/f117c7571d9739a4a5b500e3026bba769e9fe624)?

asyncio/events.py
1st1: > However I don't think it is possible to re-use the outer try-except since policy.set_running_loop might raise an exception and prevent the run to be cleaned up correctly.

Well, just put it in the right place:

```python
finally:
    self._stopping = False
    self._thread_id = None
    policy.set_running_loop(None)
    self._set_coroutine_wrapper(False)
```

Let's have just one try-finally block.

asyncio/events.py
```diff
@@ -541,6 +560,25 @@ def set_child_watcher(self, watcher):
         """Set the watcher for child processes."""
         raise NotImplementedError
 
+    # Non-abstract methods
+
+    def get_event_loop(self):
+        """Return the running event loop if any, and the default event
```
1st1: The first docstring line should be one-line sentence under 72 characters long (see https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings).  Please fix all docstrings that this PR adds/modifies.

1st1:
> @1st1 I just removed the context manager and the warnings. Should I also revert the simplification for SleepTests and TimeoutTests?

Yes, absolutely.  That "simplification" is there for a reason.

I also don't think we need `set_default_loop` and `get_default_loop` methods. Again, good ideas, but not for 3.5.2.

1st1:
>> @1st1 I just removed the context manager and the warnings. Should I also revert the simplification for SleepTests and TimeoutTests?

> Yes, absolutely. That code is there for a reason.

Just realized that with `get_event_loop` using `get_running_loop`, that code is indeed becoming meaningless. The problem is that I still want unittests to actually check that event loop is passed explicitly everywhere in asyncio...

1st1:
I think we need to add a new policy specifically for unittests (the majority of them), that would simply raise an exception in `get_event_loop` and `get_running_loop`.

vxgmichel:
> Yes, absolutely. That code is there for a reason.

I seems like it used to be there for a reason, but not anymore since `TestCase.tearDown` basically do the same thing. Anyway, it has nothing to do with this PR (the modified tests pass on the master branch as well), so I'll just revert it.

1st1:
>> Yes, absolutely. That code is there for a reason.

> I seems like it used to be there for a reason, but not anymore since TestCase.tearDown basically does the same thing. Anyway, it has nothing to do with this PR (the modified tests pass on the master branch as well), so I'll just revert it.

`TestCase.setUp` sets the current loop to `None`, so that `get_event_loop` fails with a `RuntimeError` if the loop isn't passed explicitly somewhere in the codebase.

`TestCase.tearDown` explicitly closes the event loop that was used in the tests.

Both should stay the same way.

Now, the way the updated `get_event_loop` works, is that it won't fail if `set_event_loop(None)` is called.  We need a new event policy specifically for tests, that has to be a part of this PR.

1st1:
In addition to the above: tests on Travis spit out this:

```
sys:1: ResourceWarning: gc: 12 uncollectable objects at shutdown; 
use gc.set_debug(gc.DEBUG_UNCOLLECTABLE) to list them
```

vxgmichel:
> TestCase.setUp sets the current loop to None, so that get_event_loop fails with a RuntimeError if the loop isn't passed explicitly somewhere in the codebase.

OK I finally got it, we want to make sure the optional loop argument is never silently ignored. It makes sense, but it might be a problem in the long run. For instance, coroutines such as `asyncio.sleep` will probably drop their loop argument at some point:

```python
@coroutine
def sleep(delay, result=None):
    loop = get_event_loop()
   [...]
```

So `get_event_loop()` is expected to return the running loop here, even during the tests. However, the `RuntimeError` is still very useful for objects that are not coroutines, such as locks or queues. One way to deal with this issue is to instantiate such objects outside the running loop. It turns out it is already the case for some of them (maybe all of them?), see the locks for instance: 

```python
    def test_ctor_loop(self):
        loop = mock.Mock()
        lock = asyncio.Lock(loop=loop)
        self.assertIs(lock._loop, loop)

        lock = asyncio.Lock(loop=self.loop)
        self.assertIs(lock._loop, self.loop)

    def test_ctor_noloop(self):
        asyncio.set_event_loop(self.loop)
        lock = asyncio.Lock()
        self.assertIs(lock._loop, self.loop)
```

Is that a workable solution or am I missing something?

1st1:
> For instance, coroutines such as asyncio.sleep will probably drop their loop argument at some point:

That's debatable and out of the scope of this PR ;)

> Is that a workable solution or am I missing something?

The correct solution is to create a new event loop policy for tests and update the tests to use it:

```python
class Policy:
    def get_event_loop(self):
        raise RuntimeError
```

vxgmichel:
> That's debatable and out of the scope of this PR ;)

All right!

> The correct solution is to create a new event loop policy for tests and update the tests to use it

It turns that it's quite painful to do because of cyclic imports (a `TestEventLoopPolicy` would inherit from `DefaultEventLoopPolicy` that is imported by `__init__.py` from either `unix_events` or `windows_events`. I came up with another solution, dirty but well contained:

```python
 class TestCase(unittest.TestCase):
     def set_event_loop(self, loop, *, cleanup=True):
         assert loop is not None
+        # patch policy so get_event_loop doesn't return the running loop
+        policy = events.get_event_loop_policy()
+        policy.get_event_loop = policy.get_default_loop
+        def clean_policy():
+            try:
+                del policy.get_event_loop
+            except AttributeError:
+                 pass
+        self.addCleanup(clean_policy)
         # ensure that the event loop is passed explicitly in asyncio
         events.set_event_loop(None)
         if cleanup:
             self.addCleanup(loop.close)
```

Is it acceptable? Can you think of a better solution?

1st1:
I'd add a function like this to `test_utils` (untested code):

```python
def disable_get_event_loop(test_case):
    assert isinstance(test_case, unittest.TestCase)

    policy = events.get_event_loop_policy()
    if hasattr(policy, '_patched_get_event_loop'):
        return

    def get_event_loop():
        raise RuntimeError(
            'asyncio.get_event_loop() is disabled in asyncio tests')

    def reset_event_loop_method():
        policy.get_event_loop = old_get_event_loop
        del policy._patched_get_event_loop

    old_get_event_loop = policy.get_event_loop
    policy.get_event_loop = old_get_event_loop
    policy._patched_get_event_loop = True

    test_case.addCleanup(reset_event_loop_method)
```

And then I'd add a call to it to `test_utils.TestCase.new_test_loop`, `test_utils.TestCase.set_test_loop`, and all other test cases in asyncio tests that aren't inherited from `test_utils.TestCase`.

Also, in of the earlier messages, I asked you to remove `get_default_loop` and `set_default_loop` from this PR.

vxgmichel:
I just removed `get/set_default_loop`, and added the policy patch. However I changed your example: I patched `get_running_loop` instead of  `get_event_loop` so that the behavior of `get_event_loop` is the same as before this PR.

asyncio/base_events.py
```diff
@@ -342,10 +342,15 @@ def run_forever(self):
...
+            policy = events.get_event_loop_policy()
+            policy.set_running_loop(self)
+            try:
+                while True:
+                    self._run_once()
+                    if self._stopping:
+                        break
+            finally:
+                policy.set_running_loop(None)
         finally:
```
1st1: I've left a comment somewhere in this PR on how to merge two try-finally blocks into one. Please do that.

1st1:
> I changed your example though: I patched get_running_loop instead of get_event_loop so that the behavior of get_event_loop is the same as before this PR.

Alright, that works.

1st1:
Please fix the nested try-finally blocks. Also, we need tests for `get_running_loop`.

vxgmichel:
> Please fix the nested try-finally blocks

Done! I assumed that `set_event_loop` is more likely to raise an exception than `_set_coroutine_wrapper`. Let me know if you're ok with that.

> Also, we need tests for get_running_loop.

Yes, and I guess the documentation has to be updated as well. I'll do that later this week.

asyncio/events.py
```diff
@@ -579,11 +607,27 @@ def get_event_loop(self):
...
+        """Get the running event loop for the current thread if any.
+
+        This may be None or an instance of EventLoop.
+        """
+        return self._local._running_loop
+
+    def set_running_loop(self, loop):
+        """Set the running event loop for the current thread."""
+        assert loop is None or isinstance(loop, AbstractEventLoop)
+        default_loop = self._local._loop
```
1st1: `default_loop` isn't used

asyncio/events.py
```diff
@@ -514,21 +514,43 @@ class AbstractEventLoopPolicy:
...
+        """Set the running event loop for the current context.
+
+        The loop argument can be None to clear the former running loop.
+        This method should be called by the event loop itself to set the
+        running loop when it starts, and clear it when it's done.
+        """
         raise NotImplementedError
 
     def new_event_loop(self):
         """Create and return a new event loop object according to this
```
1st1: Please fix this docstring too, while you're here.

asyncio/base_events.py
```diff
@@ -339,6 +339,8 @@ def run_forever(self):
         self._check_closed()
         if self.is_running():
             raise RuntimeError('Event loop is running.')
+        policy = events.get_event_loop_policy()
+        policy.set_running_loop(self)
```
1st1: I'd put `set_running_loop` just before `try`

asyncio/base_events.py
```diff
@@ -350,6 +352,7 @@ def run_forever(self):
             self._stopping = False
             self._thread_id = None
             self._set_coroutine_wrapper(False)
+            policy.set_running_loop(None)
```
1st1: I think we should call `policy.set_running_loop` right after `self._thread_id = None`

asyncio/base_events.py
vxgmichel: if `set_running_loop` is done after `self._thread_id = threading.get_ident()` but before `try:`, the loop itself will still be considered running if `set_running_loop` raises an exception (this exception might be legitimate for other event loop policies). So what about:

```
        policy = events.get_event_loop_policy()
        self._set_coroutine_wrapper(self._debug)
        self._thread_id = threading.get_ident()
        try:
            policy.set_running_loop(self)
            [...]
        finally:
            self._stopping = False
            self._thread_id = None
            policy.set_running_loop(None)
            self._set_coroutine_wrapper(False)
            
```

gvanrossum:
There's no way I can follow all the conversation that happened here. Why are we now collapsing get_event_loop() and get_running_loop() into one function? This seems to defeat the purpose of verifying in the unittests that there are no implicit dependencies on get_event_loop(). Should we just change all code that needs an event loop to assume get_event_loop() never raises? (E.g. sleep().)

It's possible there's a good reason, but it better be explained well, not hidden in a PR. I think if we're really changing this we should have a more public debate on the tulip list. There are a lot of words devoted to this topic in PEP 3156, and I'd hate for the PEP to become out of date like this.

asyncio/base_events.py
```diff
@@ -339,16 +339,19 @@ def run_forever(self):
         self._check_closed()
         if self.is_running():
             raise RuntimeError('Event loop is running.')
+        policy = events.get_event_loop_policy()
         self._set_coroutine_wrapper(self._debug)
         self._thread_id = threading.get_ident()
         try:
+            policy.set_running_loop(self)
```
1st1: This has to be outside of the `try` statement.

vxgmichel: If we do that, this test (not commited yet) would fail:

```python
    def test_run_loop_inside_loop(self):

        @asyncio.coroutine
        def coro():
            loop2 = asyncio.new_event_loop()
            self.assertRaises(RuntimeError, loop2.run_forever)
            self.assertFalse(loop2.is_running())  # AssertionError: True is not false
            loop2.close()

        loop = asyncio.new_event_loop()
        loop.run_until_complete(coro())
        loop.close()
```

gvanrossum: I'm beginning to think that this is all one big misunderstanding. I never
meant get_event_loop() to return a loop that's different from the running
loop. If you have multiple loops associated with the same thread you need
to implement a new EventloopPolicy that keeps track of which one is running
(e.g. via an explicit stack, or perhaps an implicit one using a local
variable to save the event loop and restoring from that in a finally
clause).

The only thing I meant to be possible is for get_event_loop() to raise an
exception, under special circumstances (especially unit tests that are
checking that no code *accidentally* relies on the default loop).

For *library* code I think there are two cases it should support:
- No default event loop, loop must always passed in
- Default event loop == current event loop, used when no loop is passed in

For *application* code I think it is totally fine to assume that
get_event_loop() returns the current, running loop. An application written
with this assumption never needs to pass a loop to something it calls
(since library code should always default to using get_event_loop()).

Note all the words devoted in the PEP to "context".

In any case I think we should stop coding and start discussing the use case
you are trying to address on the tulip list, to see whether it's real or
whether you are worrying too much (or whether maybe you should just write a
custom EventloopPolicy for your use case).

vxgmichel: @gvanrossum 
I actually don't have a use case. I simply noticed that asyncio, 3rd party libraries and users tend to pass the running loop explicitely to their coroutines, because "explicit is better than implicit". That, in my opinion, is unecessary and could be simplified.

Sadly, such a simplification conflicts with the PEP and the way asyncio unittesting currently works. I understand that those constraints makes it hard, and maybe impossible to implement. I orginally meant this PR as a proposal for a starting point towards this simplification.

However, this idea didn't seem to get a lot of support and I don't feel like trying to push it forward anymore. Maybe more people will come up with the same rationale later on, and I'll be happy to help at that time.

1st1: @gvanrossum 

> I'm beginning to think that this is all one big misunderstanding. I never
meant get_event_loop() to return a loop that's different from the running
loop. If you have multiple loops associated with the same thread you need
to implement a new EventloopPolicy that keeps track of which one is running
(e.g. via an explicit stack, or perhaps an implicit one using a local
variable to save the event loop and restoring from that in a finally
clause).

But here's the problem: if you have multiple loops associated with the same thread, policy doesn't know which one is currently running.

That's why I like the idea of adding `policy.set_running_loop` and `policy.get_running_loop` functions, because it makes policies more capable and allows you to worry less about multiple running loops event with the default policy.

gvanrossum: > if you have multiple loops associated with the same thread, policy doesn't know which one is currently running.

Then write a new policy. The whole point of having policies is that you can write different ones. The default policy does not really cater to this use case, but a policy is a really simple object.

1st1: Without `set_running_loop` and `get_running_loop` how will the policy know precisely what's going on?  All I'm saying is that `get_event_loop` and `set_event_loop` are kind of detached from the loop's lifecycle; policy doesn't receive any notification if a loop was stopped or run again.

gvanrossum: But how would you even create a second loop without the policy's
collaboration? IMO that's where it all starst. The policy should also be in
charge of deciding which loop runs -- since loops aren't created equal,
each loop has its own set of sockets it listens for (at least I would hope
so!). So switching loops should be a policy method (whose API is designed
by the specific policy -- you can't generally switch policies willy-nilly,
and libraries should never depend on the policy at all.

ambv:
Correct me if I'm wrong but it seems to me the reason of this entire debate is as simple as:
> Having to pass `loop=` around everywhere is brittle boilerplate.

We see this pattern emerging internally at Facebook because:
* we want to control the event loop during unit tests
* we sometimes run asyncio threads in a bigger threaded application

Sadly, as I said, having to pass the loop around everywhere is tiringly verbose and brittle (all it takes is for me to forget to do it once and I'm back on the default loop again). The current state of things reminds me of logging frameworks that require you to pass the logger around everywhere. Correct but brittle and inconvenient.

`get_running_loop()` is a possible way out of this. But also is Ben's "what is the current coroutine runner" proposal on async-sig, which I think is more general. I don't think there's a solid proposal for this yet but I imagine it would return an instance of asyncio.Task for asyncio, which is tied to the current running event loop.

1st1:
> get_running_loop() is a possible way out of this. But also is Ben's "what is the current coroutine runner" proposal on async-sig, which I think is more general.

I'm not so sure about the "what is the current coroutine runner" proposal.  It seems that Ben want the user to call it and depending on the result decide what to use -- `tornado.gen.multi` or `asyncio.gather`.

> I don't think there's a solid proposal for this yet but I imagine it would return an instance of asyncio.Task for asyncio, which is tied to the current running event loop.

If the task is created manually, `asyncio.Task` may get its loop reference through `get_event_loop` too.

ambv:
One case where we found the current `get_event_loop()` behavior for threading to be tricky is when you have secondary threads that schedule things on the main event loop using `call_soon_threadsafe()`. But we understood since that we STILL want `get_event_loop()` to return None in this case to ensure the user is not scheduling things on a loop that will never get run. Instead, we should only pass the bound method `loop.call_soon_threadsafe` to the secondary threads, which is both cleaner (states intent better) and makes it harder to do the wrong thing.

So... I'm warming up to the idea that we should stop worrying (and passing `loop=` around) and just embrace `get_event_loop()`.

gvanrossum:
On Mon, Jun 6, 2016 at 12:21 PM, Łukasz Langa <notifications@github.com>
wrote:

> One case where we found the current get_event_loop() behavior for
> threading to be tricky is when you have secondary threads that schedule
> things on the main event loop using call_soon_threadsafe(). But we
> understood since that we STILL want get_event_loop() to return None in
> this case to ensure the user is not scheduling things on a loop that will
> never get run. Instead, we should only pass the bound method
> loop.call_soon_threadsafe to the secondary threads, which is both cleaner
> (states intent better) and makes it harder to do the wrong thing.
>
> So... I'm warming up to the idea that we should stop worrying (and passing
> loop= around) and just embrace get_event_loop().
>

Yes!

Or write your own event loop policy (easier than you think, it's a really
simple API).

-- 

vxgmichel:
> I never meant get_event_loop() to return a loop that's different from the running
loop. [...] The policy should also be in charge of deciding which loop runs [...]

This is something I did not realize about `get_event_loop` and it does make a lot of sense. In fact, it makes the whole `get_running_loop` idea unnecessary. Maybe those two statements could be made more obvious by comparing a loop against the policy before running it:

```
    def run_forever(self):
        try:
            if asyncio.get_event_loop() != self:
                raise RuntimeError('Loop not approved by the policy')
        except asyncio.NoLoopSetError:
            # Now running in 'explicit' mode, used for unit-testing
            pass
        else:
            # The loop has been approved by the policy
            pass         
        [...]
```

The following example lists the different use cases:

```
# Typical use case
loop.run_until_complete(asyncio.sleep(1)) # OK!

# Special use case
loop = CustomLoop()
asyncio.set_event_loop(loop)
loop.run_until_complete(asyncio.sleep(1)) # OK!

# Unit-testing
loop = asyncio.test_utils.TestLoop()
asyncio.set_event_loop(None)
loop.run_until_complete(asyncio.sleep(1, loop=loop)) # OK!

# Invalid use case
loop = asyncio.new_event_loop()
loop.run_until_complete(asyncio.sleep(1, loop=loop)) # Raise!
```

gvanrossum:
Maybe. I'd be more for documenting this better though.

vxgmichel:
@1st1 I guess I can close this PR right?

By the way, a somehow [related question](http://stackoverflow.com/questions/37789420/in-which-event-loop-does-a-coroutine-run-when-using-yield-from) has been asked on stackoverflow.

1st1:
@vxgmichel Yes, let's close this PR for now. We should indeed focus on the docs; want to work on that? ;)

vxgmichel:
@1st1 I'd be happy to help, but I don't really know what is the best approach. For instance, how do we address this [comment](http://stackoverflow.com/questions/37789420/in-which-event-loop-does-a-coroutine-run-when-using-yield-from#comment63049584_37791628):
>  I'm wondering though, since explicit is better than implicit, should explicitly passing the event loop be the preferred style?

1st1:
@gvanrossum Guido, are you OK if I merge this PR? (as per https://groups.google.com/d/msg/python-tulip/yF9C-rFpiKk/tk5oA3GLHAAJ)

gvanrossum:
If you're asking do I want this feature? Yes. If you're asking me is this PR perfect? I have no idea, but I trust your review. (I don't want to have to re-understand the long discussion we had here earlier.)

1st1:
> If you're asking me is this PR perfect? I have no idea, but I trust your review.

Thank you. I'll spend some time testing the patch (with uvloop tests too) and will merge it soon.

vxgmichel:
@1st1 
Before you decide to merge this PR, I have to say I don't really like it anymore. 

A few month ago, @gvanrossum wrote:

> I never meant get_event_loop() to return a loop that's different from the running
loop. [...] The policy should also be in charge of deciding which loop runs [...]

This made me realize the whole `get_running_loop` idea is unnecessary. Instead of modifying the policy interface, we could simply compare the loop against the policy before running it: 

```python
    def run_forever(self):
        try:
            if asyncio.get_event_loop() != self:
                raise RuntimeError('Loop not approved by the policy')
        except asyncio.NoLoopSetError:
            # Now running in 'explicit' mode, used for unit-testing
            pass
        else:
            # The loop has been approved by the policy
            pass
```

Then we can advertise in the docs that `get_event_loop()` is safe to use inside coroutines. 

Would that work, or I am missing something?

asvetlov:
People may write
```
loop = asyncio.new_event_loop()
loop.run_forever()
```
Without changing default loop or even with disabling it by `asyncio.set_event_loop(None)` call.
At least I do it constantly in unit tests.

1st1:
@vxgmichel Please read the thread: https://groups.google.com/d/msg/python-tulip/yF9C-rFpiKk/tk5oA3GLHAAJ.

The point is to make `get_event_loop()` to **always** return the correct loop when it is called from a coroutine. I'm working on an overhaul of asyncio documentation right now, to deemphasize the importance of event loop in asyncio programs: don't pass it around, design APIs that don't even accept it etc.

So we want to modify **one single aspect** of `get_event_loop`: when called from coroutine always return the current event loop, no exceptions can be raised. The call must always succeed.

The point of modifying `get_event_loop` is to make all current asyncio code benefit from the change.

Since `asyncio.get_event_loop` is ultimately controlled by the policy, there should be a mechanism for event loops to notify the policy of what loop is currently running. To me, the most obvious way is to add `policy.set_running_loop` and `policy.get_running_loop`.  These must only be used by loop implementations and will be documented as such.

I like the design of the current PR.

gvanrossum:
So it'll be up to the policy to make get_event_loop() return the running
loop if one is set and the other one if not, right? And there won't be a
public interface for get_running_loop().

Still it's an API change -- policies need to implement
get/set_running_loop(). What if someone has a custom policy that doesn't
implement these and doesn't inherit from AbstractEventLoopPolicy? (It could
still be "an instance of" it, using the ABC registration API.) I wonder if
this would be a PEP change.

There could be a different way to implement this, making the "running" loop
a thread-local that is returned by the global get_event_loop() function in
preference over calling the policy's event loop.

1st1:
> There could be a different way to implement this, making the "running" loop
a thread-local that is returned by the global get_event_loop() function in
preference over calling the policy's event loop.

I like this approach more. Does this make sense to you:

1. We add `asyncio.set_current_loop()` and `asyncio.get_current_loop()` (will use a global asyncio TLS object).

2. We modify `asyncio.get_event_loop()` to first call `asyncio.get_current_loop()`. If there's no current loop, we default to `policy.get_current_loop()`.

This way there is no API change for policies, and it's easy to add support of this to third-patry loop implementations.

I think I'll close this PR and make a new one implementing this strategy.

gvanrossum:
Sounds good. Maybe name the new APIs get/set_running_loop() or
get/set_active_loop()? To emphasize that this is the loop used by all
coroutines. Also maybe these should be named _get_*() (i.e. an underscore
prefix) to clarify that they should only be called by event loop
implementations, not by users. Or at least for the _set_ variant.

asvetlov:
I vote on underscores for both getter and setter -- this API is really private.

PR #356, opened 2016-06-01
Fix typo

zware:
maybe -> may be


gvanrossum:
Honestly I think a diff just to correct one tiny typo is not worth it.

PR #357, opened 2016-06-03
WIP: Begin to allow getaddrinfo as a fallback in connect_socket.

ajdavis:
This is an approach to http://bugs.python.org/issue27136. It allows getaddrinfo for addresses we can't parse, like the Bluetooth address in that bug report. My theory is that, since getaddrinfo no longer has a mutex on popular platforms, it's ok to call it to confirm that an address has been resolved already, in cases where we can't confirm it in asyncio.

The `BaseSelectorEventLoopTests` pass with Python 3.4 and 3.5 on Mac, Linux, and Windows; other tests would need to be updated before this can be merged.

Based on @1st1's suggestions: credit is due to him, but blame mistakes on me.

### DIFF
--- a/asyncio/base_events.py
+++ b/asyncio/base_events.py
@@ -16,10 +16,8 @@ to modify the meaning of the API call itself.
 
 import collections
 import concurrent.futures
-import functools
 import heapq
 import inspect
-import ipaddress
 import itertools
 import logging
 import os
@@ -86,11 +84,12 @@ if hasattr(socket, 'SOCK_CLOEXEC'):
     _SOCKET_TYPE_MASK |= socket.SOCK_CLOEXEC
 
 
-@functools.lru_cache(maxsize=1024, typed=True)
 def _ipaddr_info(host, port, family, type, proto):
-    # Try to skip getaddrinfo if "host" is already an IP. Since getaddrinfo
-    # blocks on an exclusive lock on some platforms, users might handle name
-    # resolution in their own code and pass in resolved IPs.
+    # Try to skip getaddrinfo if "host" is already an IP. Users might have
+    # handled name resolution in their own code and pass in resolved IPs.
+    if not hasattr(socket, 'inet_pton'):
+        return
+
     if proto not in {0, socket.IPPROTO_TCP, socket.IPPROTO_UDP} or host is None:
         return None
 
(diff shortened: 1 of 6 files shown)


asyncio/base_events.py
```diff
@@ -602,17 +602,24 @@ def _getaddrinfo_debug(self, host, port, family, type, proto, flags):
...
     def getaddrinfo(self, host, port, *,
                     family=0, type=0, proto=0, flags=0):
+        if self._debug:
+            return self.run_in_executor(None, self._getaddrinfo_debug,
+                                        host, port, family, type, proto, flags)
+        else:
+            return self.run_in_executor(None, socket.getaddrinfo,
+                                        host, port, family, type, proto, flags)
+
+    def _ensure_resolved(self, host, port, *,
```
1st1: Let's make this a top-level function in base_events.py.

ajdavis: Done.

ajdavis:
Tests pass, but there's a ResourceWarning on Windows `unclosed transport <asyncio.sslproto._SSLProtocolTransport>` in `ProactorEventLoopTests.test_create_server_ssl_verify_failed` that I'm having trouble diagnosing.

Nevertheless, this is ready for review.

Here's the general idea:
* loop.create_connection resolves the host
* The resolved address is passed to sock_connect
* Once upon a time, sock_connect called getaddrinfo to check that the host was resolved (in case it was called via a different path)
* Then, I updated sock_connect to parse the host, verify it's resolved, otherwise assert
* But we discovered (issue 27136) that parsing all hosts is too hard
* So here, I fall back to using getaddrinfo if we can't parse the host, getting the best of both previous versions

Since no major platform locks around getaddrinfo any more, the original motivation for avoiding getaddrinfo is less strong now.

1st1:
Thanks, Jesse, a few notes:

1. Let's not use `ipaddress` module in `_ipaddr_info`.

2. Add `if not hasattr(socket, 'inet_pton'): return` at the top of `_ipaddr_info`

3. We no longer need `lru_cache` for `_ipaddr_info`

4. What about `loop.create_server` -- does it need to be updated?

asyncio/base_events.py
```diff
@@ -123,46 +122,26 @@ def _ipaddr_info(host, port, family, type, proto):
...
-            addr = ipaddress.IPv6Address(host.partition('%')[0])
-        except ValueError:
-            return None
-
-    af = socket.AF_INET if addr.version == 4 else socket.AF_INET6
-    if family not in (socket.AF_UNSPEC, af):
-        # "host" is wrong IP version for "family".
-        return None
+            if af == socket.AF_INET6:
+                socket.inet_pton(af, host.partition('%')[0])
```
1st1: I think we should also just return `None` if there is a `%` character in the host. The zone index should somehow influence IPv6 `scope_id`, something we can't do here.

1st1:
Thanks, Jesse. Left a couple more comments. Also, the Windows build is failing.

ajdavis:
@1st1 I think I've fixed everything you commented on? I'm proposing this patch as final.

1st1:
Pushed in ed17848b271b07bc0de8fa08d8777128ca9ca789. See also 88c0ba64e0c86f125d57915b81a748da612d78e1 and b4e5769931ce2e76dcc65696caab63ab5057cc1f.

Hope that we've finally figured this out :)  Thanks, Jesse.

ajdavis:
Thanks Yury! I appreciate learning and contributing with you.

PR #365, opened 2016-06-27
Use socket specs for getaddrinfo() in sock_connect()

Martiusweb:
In ``BaseSelectorEventLoop.sock_connect()``, an address is resolved if needed, but
``getaddrinfo()`` isn't called with the socket's parameters (family, proto, etc).
If the resolved address mismatch the socket type, the connection fails.

For instance, if ``getaddrinfo()`` returns an IPv6 (``AF_INET6``) resolution and the
socket is of family ``AF_INET``, ``sock_connect()`` eventually raises a ``TypeError``.

This PR is a follow-up to #357.


ajdavis:
LGTM, thanks.

1st1:
LGTM

1st1:
Thanks a lot!

Issue #368, opened 2016-07-03
Spurious "unclosed transport" and "Exception ignored in" message when connecting Unix pipes


ronf:
If connect_read_pipe() or connect_write_pipe() is called with an object which is not a pipe, socket, or character device, a ValueError is raised indicating that other file types are not allowed. However, the partially constructed transport object triggers an incorrect "unclosed transport" error when **del** is called on it, and another exception is raised during that call due to the "_closing" member variable having not yet been set due to the constructor never finishing. Here's a test program which reproduces this:

``` python
import asyncio

class Writer:
    pass

async def run(loop):
    f = open('/tmp/x', 'wb')

    try:
        transport, protocol = await loop.connect_write_pipe(Writer, f)
    except ValueError as exc:
        print('Exception:', exc)

loop = asyncio.get_event_loop()
loop.run_until_complete(run(loop))
loop.close()
```

Here's the output I get when running this with Python 3.5.2:

```
Exception: Pipe transport is only for pipes, sockets and character devices
Exception ignored in: <object repr() failed>
Traceback (most recent call last):
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/asyncio/unix_events.py", line 576, in __del__
    warnings.warn("unclosed transport %r" % self, ResourceWarning)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/asyncio/unix_events.py", line 456, in __repr__
    elif self._closing:
AttributeError: '_UnixWritePipeTransport' object has no attribute '_closing'
```

The first line of output is expected, but the "Exception ignored in" message is the problem.

Before Python 3.5.2, this would show up as an "Exception ignored in " message with no traceback, due to another exception raised when calling **repr**() on this object. This is also related to _closing not having been set yet.

It would be good to set always set _closing before any exception is raised, and probably not set _pipe until after this, so the "unclosed transport" message won't be triggered in __del__().

gvanrossum:
Do you feel up to making a PR?

--Guido (mobile)

ronf:
Sure - I'll try and put something together for this later this week.

1st1:
Closing this one since the relevant PR has been merged.

Issue #369, opened 2016-07-03
connect_write_pipe doesn't work properly on fds that support bidirectional I/O


ronf:
The connect_read_pipe() and connect_write_pipe() functions claim to work on character devices, sockets, and pipes. However, _UnixWritePipeTransport assumes that a write pipe has closed if _read_ready() ever triggers on it. This won't work for TTYs that allow both input & output on the same file descriptor. I think the same problem will occur with sockets.

I've been able to reproduce this failure by passing in sys.stdout to connect_write_pipe() when it is not redirected and pointing at /dev/tty. In that case, any input on stdin immediately called the write pipe opened on stdout to close.

Here's code which illustrates this:

```python
import asyncio
import logging
import sys

logging.basicConfig(level='DEBUG')

class Reader(asyncio.Protocol):
    def data_received(self, data):
        print('Got data:', data)

class Writer(asyncio.BaseProtocol):
    pass

async def run(loop):
    stdin_t, _ = await loop.connect_read_pipe(Reader, sys.stdin)
    stdout_t, _ = await loop.connect_write_pipe(Writer, sys.stdout)

loop = asyncio.get_event_loop()
loop.run_until_complete(run(loop))
```

When run with PYTHONASYNCIODEBUG=1, I see the following output:

```
DEBUG:asyncio:Using selector: KqueueSelector
DEBUG:asyncio:Read pipe 0 connected: (<_UnixReadPipeTransport fd=0 polling>, <__main__.Reader object at 0x103f62828>)
DEBUG:asyncio:Write pipe 1 connected: (<_UnixWritePipeTransport fd=1 idle bufsize=0>, <__main__.Writer object at 0x1046d9208>)
a    <--- Input provided here
INFO:asyncio:poll took 2221.794 ms: 2 events
Got data: b'a\n'
INFO:asyncio:<_UnixWritePipeTransport fd=1 idle bufsize=0> was closed by peer
```

gvanrossum:
Can you suggest a fix, or do we have to update the docs?

--Guido (mobile)

ronf:
If we can find a different way to detect when a "write" pipe has been closed that doesn't rely on it being marked readable, it may be possible to continue to support sockets and other character devices. I'll experiment a bit when I get a chance and post my results here, or submit a PR if I can come up with a clean fix.

ronf:
[...]

PR #370, opened 2016-07-04
Fix ordering issues in UNIX read/write pipe transport constructors

ronf:
Here's a pull request designed to fix the issue reported in #368.

This commit re-orders the initialization code in the _UnixReadPipeTransport
and _UnixWritePipeTransport constructors to make sure all members are
assigned a value, even in the case where ValueError is raised due to
an incompatible type of pipe. This avoids exceptions being raised in
__repr__() due to the missing members.

In the case where ValueError is raised, this commit also clears the values
of _pipe, _fileno, and _protocol since this transport isn't returned,
avoiding a spurious "unclosed transport" warning when the object is
garbage-collected.

### DIFF
--- a/asyncio/base_events.py
+++ b/asyncio/base_events.py
@@ -707,8 +707,6 @@ class BaseEventLoop(events.AbstractEventLoop):
             raise ValueError(
                 'host and port was not specified and no sock specified')
 
-        sock.setblocking(False)
-
         transport, protocol = yield from self._create_connection_transport(
             sock, protocol_factory, ssl, server_hostname)
         if self._debug:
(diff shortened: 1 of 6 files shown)


ronf:
Sorry - I accidentally checked in a proposed fix to issue #369 into this same pull request. I should have used a separate branch for that. Let me know if you'd like me to split them.

1st1:
Hi Ron,

> Sorry - I accidentally checked in a proposed fix to issue #369 into this same pull request. I should have used a separate branch for that. Let me know if you'd like me to split them.

Yes, please split them. That will make the review easier.

gvanrossum:
The git-level fix should be easy right? `git reset HEAD^; git push -f`

ronf:
Thanks - done. I'll submit the other patch in a separate pull request.

1st1:
Should we add a unittest checking that the repr works after a ValueError?

ronf:
This is a bit tricky to do as a black-box test, as the object is not returned to the caller when a ValueError is raised, so there's nothing to call repr on from within the unit test code.

ronf:
I tried to rebase this to fix a conflict with the merge of #374, but it didn't quite come out right. I'm going to close this and submit a new pull request with an updated fix for #368.

Issue #286, opened 2015-11-12
Add new TCPTransport base class


1st1:
In [uvloop](https://github.com/1st1/uvloop) (more details [here](https://github.com/python/asyncio/issues/282#issuecomment-155957235)) I implement Transports on top of libuv streams. Here's documentation on [TCP](http://docs.libuv.org/en/v1.x/tcp.html#c.uv_tcp_nodelay) and [general streams](http://docs.libuv.org/en/v1.x/stream.html#api).

Streams are an opaque abstraction over sockets API. All buffering is implemented in the libuv core. There is actually no way to access the underlying socket, which causes some problems. Right now, if you want to set `TCP_NODELAY` or `SO_KEEPALIVE`, the only way is to call `Transport.get_extra_info('socket')` and set them manually.  That's what aiohttp is [doing](https://github.com/KeepSafe/aiohttp/blob/be9ab0e993785ad94bccfd9e4d927858c75bf725/aiohttp/server.py#L30), for instance.

I propose to add a new `Transport` base class - `TCPTransport`. It will have two specific to TCP methods: `set_nodelay(bool enabled)` and `set_keepalive(bool enabled)`.

Without these methods, it's virtually impossible for me to create a transport that provides this functionality on top of libuv.  This is very similar to how this is done in [Twisted](http://twistedmatrix.com/documents/8.1.0/api/twisted.internet.tcp.Connection.html) and in [Tornado](http://www.tornadoweb.org/en/stable/iostream.html#tornado.iostream.BaseIOStream.set_nodelay).

haypo:
"There is actually no way to access the underlying socket, which causes some problems. "

Hum, how do you plan to set TCP_NODELAY flag if you don't have access to the socket?

There are a *lot* of socket flags and options. I'm not sure that it's a good idea to add a method for each flag. It will be difficult to document the available of each method depending on the platform and the implementation of the event loop. We already have to take care in the documentation of subtle differences between SelectorEventLoop on UNIX and on Windows, and the special case of ProactorEventLoop.

asvetlov:
Thinking on the issue I would suggest another solution: don't export `'socket'` extra info by uvloop (it's impossible) but add uv-specific slot for controlling socket options in libuv way.

[...]

PR #373, opened 2016-07-05
Set TCP_NODELAY on TCP transports by default

1st1:
This PR enables `TCP_NODELAY` for all TCP transports.

### DIFF
--- a/asyncio/selector_events.py
+++ b/asyncio/selector_events.py
@@ -39,6 +39,17 @@ def _test_selector_event(selector, fd, event):
         return bool(key.events & event)
 
 
+if hasattr(socket, 'TCP_NODELAY'):
+    def _set_nodelay(sock):
+        if (sock.family in {socket.AF_INET, socket.AF_INET6} and
+                sock.type == socket.SOCK_STREAM and
+                sock.proto == socket.IPPROTO_TCP):
+            sock.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
+else:
+    def _set_nodelay(sock):
+        pass
+
+
 class BaseSelectorEventLoop(base_events.BaseEventLoop):
     """Selector event loop.
 
@@ -639,6 +650,11 @@ class _SelectorSocketTransport(_SelectorTransport):
         super().__init__(loop, sock, protocol, extra, server)
         self._eof = False
         self._paused = False
+
+        # Disable the Nagle algorithm -- small writes will be
+        # sent without waiting for the TCP ACK.  This generally
+        # decreases the latency (in some cases significantly.)
+        _set_nodelay(self._sock)
 
         self._loop.call_soon(self._protocol.connection_made, self)
         # only start reading when connection_made() has been called


gvanrossum:
Shouldn't we at least have a way to turn it off? Or why do sockets not have this on by default?

jimfulton:
Assuming that travis eventually goes green, LGTM. Thanks!

1st1:
@gvanrossum Please see https://github.com/python/asyncio/issues/286, where I wanted to add another base Transport class - `TCPTransport`, with a `set_nodelay` method.  I still think it'd be a good idea ;)

> Or why do sockets not have this on by default?

NODELAY can be a bad thing in cases where you want to send as few packages as possible. Setting it causes writes to be sent asap, instead of waiting until the buffer is full or for a TCP ACK.

Without TCP_NODELAY socket operations can have a huge latency. It's particularly important to set for database drivers, HTTP servers and clients etc, where we want the latency to be as small as possible. For modern applications always setting TCP_NODELAY is a good default (and, for instance, that's what Golang does for all TCP connections).

Martiusweb:
If TCP_NODELAY is enabled, a call to write()/send() means that a TCP packet is emitted regardless of if the client ACK'ed the previous packet (as long as the congestion window isn't full), while when disabled, the packet is emited only when there is as much as the MTU to send (or after a well known delay, usually 200ms).

This is a full win when the user sends a whole message (as seen by the application protocol) in one call, and for "synchronized" protocols such as http when peers don't usually write on the socket "at the same time".

Regarding the PR, isn't TCP_NODELAY available on windows/with proactor?

1st1:
> If TCP_NODELAY is enabled, a call to write()/send() means that a TCP packet is emitted regardless of if the client ACK'ed the previous packet (as long as the congestion window isn't full), while when disabled, the packet is emited only when there is as much as the MTU to send (or after a well known delay, usually 200ms).

I'm curious what happens when you use [os.writev()](https://docs.python.org/3/library/os.html#os.writev) call – writing a bunch of buffers with one syscall.  Will a bunch of small buffers sent with `writev` be aggregated in one TCP packet?

> Regarding the PR, isn't TCP_NODELAY available on windows/with proactor?

Yes, good catch. I'll add it to the proactor too.

1st1:
>> If TCP_NODELAY is enabled, a call to write()/send() means that a TCP packet is emitted regardless of if the client ACK'ed the previous packet (as long as the congestion window isn't full), while when disabled, the packet is emited only when there is as much as the MTU to send (or after a well known delay, usually 200ms).

>I'm curious what happens when you use os.writev() call – writing a bunch of buffers with one syscall. Will a bunch of small buffers sent with writev be aggregated in one TCP packet?

I think I found the answer -- `writev` gathers the output and transfers the data in a single operation. This is important because we want to start using it when https://github.com/python/asyncio/pull/339 is merged.

Martiusweb:
Only one packet is sent, even with TCP_NODELAY (tested with Linux 4.6.3 and wireshark).

asyncio/selector_events.py
```diff
@@ -634,6 +645,11 @@ def __init__(self, loop, sock, protocol, waiter=None,
         self._eof = False
         self._paused = False
 
+        # Disable the Nagle algorithm -- small writes will be
+        # sent without waiting for the TCP ACK.  This generally
+        # decreases the latency (in some cases significantly.)
+        _set_nodelay(self._sock)
```
socketpair: This will work only for TCP sockets, and not for UNIX stream sockets...

1st1: Good catch. We need to add functional unittests for TCP/UNIX transports to catch these sort of errors...

socketpair: Also, accepted sockets should be set as NODELAY too. Please check that this is true.

asyncio/selector_events.py
```diff
@@ -39,6 +39,17 @@ def _test_selector_event(selector, fd, event):
         return bool(key.events & event)
 
 
+if hasattr(socket, 'TCP_NODELAY'):
+    def _set_nodelay(sock):
+        try:
+            sock.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
+        except OSError:
+            pass
```
socketpair: 1. Hiding error is bad practice. Is any real case when it will fail?
2. Why not to embed that code into place where it is called ?

asyncio/selector_events.py
```diff
@@ -39,6 +39,17 @@ def _test_selector_event(selector, fd, event):
         return bool(key.events & event)
 
 
+if hasattr(socket, 'TCP_NODELAY'):
```
socketpair: Is there case when this option is not available ? If yes, please add comment to source code. If no, just remove that check.

asyncio/selector_events.py
1st1: IIRC some OSes have TCP_NODELAY defined, but not implemented in the kernel, in which case you'll get an OSError.

asyncio/selector_events.py
1st1: Yes, some old linux kernels don't implement this. And some other OSes might not too.

How would a comment add any kind of clarity here?

asyncio/selector_events.py
socketpair: Can you name that OSes ?

asyncio/selector_events.py
socketpair: 1. Seems, you mean TCP_CORK, which was intorduced near Linux 2.4. AFAIK, TCP_NODELAY was implemented much earlier
2. Which OSes really still not implements that?

Don't misunderstand me. I'm asking that in order to prevent code bloating and helping the code to be simple to read and maintain.

asyncio/selector_events.py
1st1: Without extensive googling no. Golang ignores any errors when setting TCP_NODELAY for similar reasons.

It's an optional flag, raising an error doesn't make any sense here as it can break a perfectly working program in some obscure case. What I can do is to log it in debug mode, but I don't think we need that.

asyncio/selector_events.py
1st1: > Seems, you mean TCP_CORK, which was intorduced near Linux 2.4. AFAIK, TCP_NODELAY was implemented much earlier

Maybe you're right. Anyways, from `socketmodule.c`:

```c
#ifdef  TCP_NODELAY
    PyModule_AddIntMacro(m, TCP_NODELAY);
#endif
```

CPython doesn't make an assumption that it's always available, so shouldn't we.

asyncio/selector_events.py
1st1: Alright. I'll prepare a new patch some time later. Thanks for reviewing this iteration!

asyncio/selector_events.py
socketpair: https://www.python.org/dev/peps/pep-0011/#no-longer-supported-platforms

asyncio/selector_events.py
socketpair: See also:
https://github.com/KeepSafe/aiohttp/issues/664

socketpair: > Good catch. We need to add functional unittests for TCP/UNIX transports to catch these sort of errors...

Tests probably shows nothing, since error is swallowed. Or, tests should test that TCP_NODELAY has been actually set.

1st1:
@socketpair I don't care that much about "performance" here. Questions like "what is faster to build, set or tuple" should only be addressed when you have an extremely tight and performance critical loop in your algorithm. Even then your decision will be very CPython specific. Anyways, I like the code as it is now.

SethMichaelLarson:
@1st1 Sorry, misunderstood the usage of _set_nodelay, I thought it was applied to each socket created by asyncio, not just server sockets. If that's the case then how it is now is fine.

1st1:
> @1st1 Sorry, misunderstood the usage of _set_nodelay, I thought it was applied to each socket created by asyncio, not just server sockets. If that's the case then how it is now is fine.

@SethMichaelLarson @socketpair Hm, `NODELAY` should be applied to client connections & server connections created with `loop.create_connection` & `loop.create_server`. Transports created by those functions are instances of `_SelectorSocketTransport` which applies `NODELAY` in its constructor, so all of them have the flag set. Am I missing something?

SethMichaelLarson:
@1st1 Oh, so it applies to all sockets as I thought previously. Disregard my above comment. If that's the case then I think I'm still +1 on not using a tuple/set to check socket.family just because of how often it's used.

asvetlov:
I believe the check for set is nothing comparing to syscall.
But using `TCP_NODELAY` makes sense for almost all use cases and should be enabled by default.

AndreLouisCaron:
@1st1 

> Yes, good catch. I'll add it to the proactor too.

I just installed Python 3.6 and seems like the proactor doesn't use `TCP_NODELAY`.  Was there some kind of limitation that prevented adding it?  If it's a simple omission, would you be open to a PR that adds it there too?

AndreLouisCaron:
For context, I opened issue aio-libs/aiomysql#149 because it has its own switch for `TCP_NODELAY` (presumably for Python 3.4 and 3.5) that breaks when run against the proactor event loop.

It would be neat if we could have this fixed upstream and if we can remove that broken code from aiomysql.

asyncio/selector_events.py
```diff
@@ -640,6 +651,11 @@ def __init__(self, loop, sock, protocol, waiter=None,
         self._eof = False
         self._paused = False
 
+        # Disable the Nagle algorithm -- small writes will be
+        # sent without waiting for the TCP ACK.  This generally
```
socketpair: No, TCP_NODELAY is not connected with ACK. Actually, when nodelay is NOT set, kernel will delay sending small TCP packet, waiting for possible additional data to form one big packet.

methane: @socketpair see https://en.wikipedia.org/wiki/Nagle%27s_algorithm
nagle waits ACK or enough data in send buffer.  That's why delayed ACK + nagle cause trouble.

socketpair: image

You have said only part of algorightm (that is connected with ACK). Much more important thing is what I tried to describe (that is connected with writes < MSS).

socketpair: P.S. I'm not right too. It's best not to describe Nagle's algorithm in the comment.

PR #378, opened 2016-07-09
Provide an API to handle connections that are accepted outside of ayncio

jimfulton:
This addresses: https://bugs.python.org/issue27392

### DIFF
--- a/asyncio/base_events.py
+++ b/asyncio/base_events.py
@@ -980,6 +980,36 @@ class BaseEventLoop(events.AbstractEventLoop):
         return server
 
     @coroutine
+    def handle_connection(self, protocol_factory, sock, ssl=None):
+        """Handle an accepted connection
+
+        This is used by servers that accept connections outside of
+        asyncio but that use asyncio to handle connections.
+
+        This method is a coroutine.  When completed, the coroutine
+        returns a (transport, protocol) pair.
+        """
+        sock.setblocking(False)
+
+        protocol = protocol_factory()
+        waiter = self.create_future()
+        if ssl:
+            sslcontext = None if isinstance(ssl, bool) else ssl
+            transport = self._make_ssl_transport(
+                sock, protocol, sslcontext, waiter,
+                server_side=True)
+        else:
+            transport = self._make_socket_transport(sock, protocol, waiter)
+
+        try:
+            yield from waiter
+        except:
+            transport.close()
+            raise
+
+        return transport, protocol
+
+    @coroutine
     def connect_read_pipe(self, protocol_factory, pipe):
         protocol = protocol_factory()
         waiter = self.create_future()
--- a/asyncio/base_events.py
+++ b/asyncio/base_events.py
@@ -981,8 +981,8 @@ class BaseEventLoop(events.AbstractEventLoop):
         return server
 
     @coroutine
-    def connect_accepted_socket(self, protocol_factory, sock, ssl=None):
-        """Handle an accepted connection
+    def connect_accepted_socket(self, protocol_factory, sock, *, ssl=None):
+        """Handle an accepted connection.
 
         This is used by servers that accept connections outside of
         asyncio but that use asyncio to handle connections.
--- a/asyncio/base_events.py
+++ b/asyncio/base_events.py
@@ -707,8 +707,6 @@ class BaseEventLoop(events.AbstractEventLoop):
             raise ValueError(
                 'host and port was not specified and no sock specified')
 
-        sock.setblocking(False)
-
         transport, protocol = yield from self._create_connection_transport(
             sock, protocol_factory, ssl, server_hostname)
         if self._debug:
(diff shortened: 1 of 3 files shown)


1st1:
I don't like the name -- how about `wrap_socket()`?

jimfulton:
``wrap_socket`` implies that it's for wrapping client or server sockets, but it's not. It's only for handling server sockets.  Also, I prefer a name that reflects goal, not mechanism.

I think the name should be discussed over here: https://bugs.python.org/issue27392.

gvanrossum:
What's wrong with the Python 3.4 tests on Windows (AppVeyor)?

Can we share more code between this and _create_connection_transport()?

I'll deliberate the name in the Python tracker issue.

jimfulton:
I have no idea WRT Python 3.4 on windows. It seems to be hanging. I don't have a windows development environment.  Do you by any chance know where I can find a good howto for setting one up?

jimfulton:
Just a note WRT the name: Really it's about more than the name. @1st1 argues that there should be a new API for wrapping already-connected sockets, for both clients and servers, and that create_connection shouldn't accept connected sockets. I think this is a pretty good argument.

jimfulton:
WRT sharing code, I guess I could make ``_create_connection_transport`` handle both the client and server case.  I'll do that.

jimfulton:
OK, more sharing. Maybe I'll be lucky and the windows tests will pass. :)

gvanrossum:
It's also about more than finding the perfect API. (Not that everyone would
even agree on that.)

There are several APIs that create and configure a socket in a certain way
and wrap it in a transport/protocol pair, and all of them have the option
of taking an existing socket. I think that's a reasonable model and we
should not abandon it.

The complication we have here is that there are two types of server
sockets, the "listening" kind and the "connected" kind. If you're really
into finding the perfect API, maybe the right thing to do would be to
search for a way to refactor create_server()?

asyncio/base_events.py
```diff
@@ -980,6 +981,25 @@ def create_server(self, protocol_factory, host=None, port=None,
         return server
 
     @coroutine
+    def connect_accepted_socket(self, protocol_factory, sock, ssl=None):
```
1st1: Let's make `ssl` a keyword-only argument (as it is in `create_connection` and `create_server`).

asyncio/base_events.py
```diff
@@ -980,6 +981,25 @@ def create_server(self, protocol_factory, host=None, port=None,
...
+        """Handle an accepted connection
```
1st1: Please put a period at the end of the first line (to silence linters).

1st1:
Aside from a couple of nits it looks good.

jimfulton:
The tests hanging for the ProactorEventLoop with SSL.  I'll skip that case.

BTW, running tests verbosely is helpful for cases like this. Is there any harm in always running the appveyor tests verbosely?

1st1:
> The tests hanging for the ProactorEventLoop with SSL. I'll skip that case.

Strange, sounds like we need to open an issue for this (but it shouldn't block this PR).

> BTW, running tests verbosely is helpful for cases like this. Is there any harm in always running the appveyor tests verbosely?

Yes, it makes sense. Can you make a separate PR to run tests in verbose mode both for Travis and AppVeyor?

jimfulton:
>
> The tests hanging for the ProactorEventLoop with SSL. I'll skip that case.
>
> Strange, sounds like we need to open an issue for this (but it shouldn't
> block this PR).
>

No, It's documented:
https://docs.python.org/3/library/asyncio-eventloops.html#windows

> BTW, running tests verbosely is helpful for cases like this. Is there any
> harm in always running the appveyor tests verbosely?
>
> Yes, it makes sense. Can you make a separate PR to run tests in verbose
> mode both for Travis and AppVeyor?
>
Sure.

Jim

-- 
Jim Fulton
http://jimfulton.info

jimfulton:
OK, all tests passing. Assuming this is merged, I'll update the docs. Where do I do that?

jimfulton:
OK, all tests passing. Assuming this is merged, I'll update the docs. Where do I do that?

1st1:
LGTM. Will merge it in a few hours.

> OK, all tests passing. Assuming this is merged, I'll update the docs. Where do I do that?

@gvanrossum Should we copy asyncio docs from CPython to this repo and update the `update_stdlib.sh` script to handle that?

gvanrossum:
Regarding the docs: that would only be useful if we also had the machinery
to build the docs here.

I think it may be time actually to give up on this github repo and focus
all development in the CPython repo. Especially since I also think that
asyncio should drop the "provisional" status in Python 3.6.

Maybe we could keep this repo for something called "backports/asyncio",
more in line with the backports versions that exist of other Python 3
packages (e.g. enum34). That might also make it easier to distribute an
alternative version that can override the stdlib version.

jimfulton:
Can this be merged? If it is, will it be in 3.6? I'm happy to update the docs wherever.

1st1:
> Can this be merged? If it is, will it be in 3.6? I'm happy to update the docs wherever.

Merged.  Yes, it will be in 3.6 (and I'll add it to uvloop in a couple of days).

For the docs please upload a patch to https://bugs.python.org/issue27392.

@gvanrossum 

> I think it may be time actually to give up on this github repo and focus
all development in the CPython repo. Especially since I also think that
asyncio should drop the "provisional" status in Python 3.6.

Please let's wait until CPython migrates to GitHub. I like code review workflow here much better.

Issue #390, opened 2016-07-27
AttributeError: 'NoneType' object has no attribute 'get_debug'


lphuberdeau:
Writing unit tests for asyncio code, I have a few tests that fail occasionally with this error. It seems to be timing-related.

I use [loop_context()](https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/test_utils.py#L502) from aiohttp to make sure all tests are isolated.

The problem is essentially that _loop is not set by the time logging is attempted, which may be related to the loop closing and unregistering after wait_for() canceled the job.

```
Traceback (most recent call last):
  File "/../tests/fixtures.py", line 24, in wrapper
    loop.run_until_complete(f(*args, loop=loop, **kwargs))
  File "/usr/lib/python3.5/asyncio/base_events.py", line 387, in run_until_complete
    return future.result()
  File "/usr/lib/python3.5/asyncio/futures.py", line 274, in result
    raise self._exception
  File "/usr/lib/python3.5/asyncio/tasks.py", line 239, in _step
    result = coro.send(None)
...
  File "....py", line 48, in ls
    return await asyncio.wait_for(self.read_lines(command), 30.0, loop=self.loop)
  File "/usr/lib/python3.5/asyncio/tasks.py", line 392, in wait_for
    return fut.result()
  File "/usr/lib/python3.5/asyncio/futures.py", line 274, in result
    raise self._exception
  File "/usr/lib/python3.5/asyncio/tasks.py", line 239, in _step
    result = coro.send(None)
  File "....py", line 58, in read_lines
    stdin=asyncio.subprocess.DEVNULL
  File "/usr/lib/python3.5/asyncio/subprocess.py", line 212, in create_subprocess_exec
    stderr=stderr, **kwds)
  File "/usr/lib/python3.5/asyncio/base_events.py", line 1079, in subprocess_exec
    bufsize, **kwargs)
  File "/usr/lib/python3.5/asyncio/unix_events.py", line 187, in _make_subprocess_transport
    self._child_watcher_callback, transp)
  File "/usr/lib/python3.5/asyncio/unix_events.py", line 808, in add_child_handler
    self._do_waitpid(pid)
  File "/usr/lib/python3.5/asyncio/unix_events.py", line 841, in _do_waitpid
    if self._loop.get_debug():
AttributeError: 'NoneType' object has no attribute 'get_debug'

```

In this case, the culprit is this one, but the pattern repeats often in the code.

 ```python
            if self._loop.get_debug():
[...]

PR #391, opened 2016-08-02
 Refuse handlers if the child watcher has no loop attached

vxgmichel:
This PR follows up on the discussion from issue #390:

- f348122 fixes the issue
- f7a5259 is a proposal and can easily be discarded

This has been tested against the [example](https://github.com/python/asyncio/issues/390#issuecomment-235871755) from issue #390.

### DIFF
--- a/asyncio/unix_events.py
+++ b/asyncio/unix_events.py
@@ -726,6 +726,7 @@ class BaseChildWatcher(AbstractChildWatcher):
 
     def __init__(self):
         self._loop = None
+        self._callbacks = {}
 
     def close(self):
         self.attach_loop(None)
@@ -738,6 +739,12 @@ class BaseChildWatcher(AbstractChildWatcher):
 
     def attach_loop(self, loop):
         assert loop is None or isinstance(loop, events.AbstractEventLoop)
+
+        if self._loop is not None and loop is None and self._callbacks:
+            warnings.warn(
+                'A loop is being detached '
+                'from a child watcher with pending handlers',
+                RuntimeWarning)
 
         if self._loop is not None:
             self._loop.remove_signal_handler(signal.SIGCHLD)
@@ -786,10 +793,6 @@ class SafeChildWatcher(BaseChildWatcher):
     This is a safe solution but it has a significant overhead when handling a
     big number of children (O(n) each time SIGCHLD is raised)
     """
-
-    def __init__(self):
-        super().__init__()
-        self._callbacks = {}
 
     def close(self):
         self._callbacks.clear()
@@ -871,7 +874,6 @@ class FastChildWatcher(BaseChildWatcher):
     """
     def __init__(self):
         super().__init__()
-        self._callbacks = {}
         self._lock = threading.Lock()
         self._zombies = {}
         self._forks = 0
--- a/asyncio/unix_events.py
+++ b/asyncio/unix_events.py
@@ -802,6 +802,11 @@ class SafeChildWatcher(BaseChildWatcher):
         pass
 
     def add_child_handler(self, pid, callback, *args):
+        if self._loop is None:
+            raise RuntimeError(
+                "Cannot add child handler, "
+                "the child watcher does not have a loop attached")
+
         self._callbacks[pid] = (callback, args)
 
         # Prevent a race condition in case the child is already terminated.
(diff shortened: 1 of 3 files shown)


1st1:
Will push this before b2.

tests/test_subprocess.py
```diff
@@ -433,6 +433,11 @@ def kill_running():
         # the transport was not notified yet
         self.assertFalse(killed)
 
+        # Clear the handlers for FastChildWatcher to avoid a warning.
+        # Is that a bug?
+        if sys.platform != 'win32':
+            asyncio.get_child_watcher()._callbacks.clear()
```
1st1: This is something new... Is there any reason you do this as part of this PR?

tests/test_unix_events.py
```diff
@@ -1396,7 +1397,11 @@ def test_set_loop_race_condition(self, m):
         with mock.patch.object(
                 old_loop, "remove_signal_handler") as m_remove_signal_handler:
 
-            self.watcher.attach_loop(None)
```
1st1: How about using `with self.assertWarnsRegexp`?

asyncio/unix_events.py
```diff
@@ -898,6 +905,12 @@ def __exit__(self, a, b, c):
 
     def add_child_handler(self, pid, callback, *args):
         assert self._forks, "Must use the context manager"
+
+        if self._loop is None:
```
1st1: Can we add a test for this?

tests/test_subprocess.py
vxgmichel: Yes, otherwise a warning from  f7a5259 is generated. Weirdly enough it only happens with the `FastChildWatcher`.  There might be a better way to address this issue though.

tests/test_unix_events.py
vxgmichel: Done, it's much cleaner now!

asyncio/unix_events.py
vxgmichel: Done!

tests/test_subprocess.py
vxgmichel: I just had a deeper look and we can get rid of this addition by fixing `FastChildWatcher` (using [some `SafeChildWatcher` code](https://github.com/python/asyncio/blob/master/asyncio/unix_events.py#L848-L854)):
```
diff --git a/asyncio/unix_events.py b/asyncio/unix_events.py
index 54cfc75..5ffa645 100644
--- a/asyncio/unix_events.py
+++ b/asyncio/unix_events.py
@@ -937,7 +937,15 @@ def _do_waitpid_all(self):
                 pid, status = os.waitpid(-1, os.WNOHANG)
             except ChildProcessError:
                 # No more child processes exist.
-                return
+                if not self._callbacks:
+                    return
+                # Some child processes are already reaped
+                # (may happen if waitpid() is called elsewhere).
+                pid = next(iter(self._callbacks))
+                returncode = 255
+                logger.warning(
+                    "Unknown child process pid %d, will report returncode 255",
+                    pid)
             else:
                 if pid == 0:
                     # A child process is still alive.
```
This also allows us to get rid of some complexity in `test_sigchld_child_reaped_elsewhere`: 
```
diff --git a/tests/test_unix_events.py b/tests/test_unix_events.py
index 6cf4417..b4a6d80 100644
--- a/tests/test_unix_events.py
+++ b/tests/test_unix_events.py
@@ -1318,12 +1318,7 @@ def test_sigchld_child_reaped_elsewhere(self, m):
         with self.ignore_warnings:
             self.watcher._sig_chld()
 
-        if isinstance(self.watcher, asyncio.FastChildWatcher):
-            # here the FastChildWatche enters a deadlock
-            # (there is no way to prevent it)
-            self.assertFalse(callback.called)
-        else:
-            callback.assert_called_once_with(58, 255)
+        callback.assert_called_once_with(58, 255)
 
     @waitpid_mocks
     def test_sigchld_unknown_pid_during_registration(self, m):
```

All the tests run OK after those modifications. Let me know which solution you prefer.

1st1: Not so sure about this solution. Let's keep the current hack-ish code.

1st1:
@asvetlov Could you please help me with reviewing this PR?

asvetlov:
Looks like I  should update my pytest fixtures for `pytest-aiohttp` by explicitly attaching/detaching test loop.
It's really nice catch. By writing fixtures for aiohttp I totally forgot about subprocess quirks on UNIX.
Honestly I almost never run subprocesses in my asyncio-based code but raising an explicit error is much better than just hanging in worst case.
Casual usage without default loop disabling is not affected as I see.

1st1:
@asvetlov Andrew, I'm on vacation, and I can't commit this myself right now. If you have time, could you please commit this pr (and merge it to cpython 3.5/3.6/default)?

1st1:
Committed by hand. Thanks, Vincent!

1st1:
See also http://bugs.python.org/issue28368

Issue #394, opened 2016-08-04
Memory footprint StreamReader.readexactly


frensjan:
```StreamReader.readexactly``` is a very handy method to implement reading frames ('chunks' of data preceded with the size of the 'chunk') from a stream. E.g.:
```python
frame_len = yield from reader.readexactly(4) # when using 32 bits sizes
frame = yield from reader.readexactly(frame_len)
```

---

However, it's memory footprint isn't ideal. A list of 'blocks' is kept until all data is read which is then joined with ```b''.join(blocks)```. This causes each 'frame' to reside in memory twice until the ```blocks``` list and the ```bytes``` objects it points to are freed (when leaving the ```readexactly``` method). So it's peak memory usage is ```O(2n)```.

For small 'frames' the only issue probably is some copying which could be avoided, but for larger 'frames' this hurts. E.g. when sending large files for processing to a machine which are larger than ```memory / #receivers / 2```.

---

A possible solution might be to use a ```bytearray``` to gather the data incrementally. E.g. something along the lines of:

```python
@asyncio.coroutine
def readexactly(self, n):
    if self._exception is not None:
        raise self._exception
     
    if not n:
        return b''
     
    if len(self._buffer) >= n and not self._eof:
        data = self._buffer[:n]
        del self._buffer[:n]
        self._maybe_resume_transport()
        return data
     
    data = bytearray(n)
    pos = 0
     
    while n:
        if not self._buffer:
            yield from self._wait_for_data('read')
 
        if self._eof or not self._buffer:
            raise asyncio.IncompleteReadError(data[:pos], pos + n)
         
        available = len(self._buffer)
        if available <= n:
            data[pos:pos + available] = self._buffer
            self._buffer.clear()
            n -= available
            pos += available
        else:
            data[pos:pos + n] = self._buffer[:n]
            del self._buffer[:n]
            n = 0
        self._maybe_resume_transport()
     
    return data
```

This implementation would have a peak memory footprint of ```O(n)```. Or ```O(n + max(map(len, blocks)))```, not sure what the other constant factors might be, probably lots more lower in the stack.

Note that the implementation above also does away with some unnecessary copying by duplicating some code from ```StreamReader.read```.
Note as well that the implementation has a short path for ```readexactly``` when the required data is available in the ```_buffer``` of the ```StreamReader```.

[...]

PR #395, opened 2016-08-05
Speedup readexactly

socketpair:
See #394

### DIFF
--- a/asyncio/streams.py
+++ b/asyncio/streams.py
@@ -448,6 +448,7 @@ class StreamReader:
         assert not self._eof, '_wait_for_data after EOF'
 
         # Waiting for data while paused will make deadlock, so prevent it.
+        # This is essential for readexactly(n) for case when n > self._limit.
         if self._paused:
             self._paused = False
             self._transport.resume_reading()
@@ -658,25 +659,22 @@ class StreamReader:
         if n == 0:
             return b''
 
-        # There used to be "optimized" code here.  It created its own
-        # Future and waited until self._buffer had at least the n
-        # bytes, then called read(n).  Unfortunately, this could pause
-        # the transport if the argument was larger than the pause
-        # limit (which is twice self._limit).  So now we just read()
-        # into a local buffer.
+        while len(self._buffer) < n:
+            if self._eof:
+                incomplete = bytes(self._buffer)
+                self._buffer.clear()
+                raise IncompleteReadError(incomplete, n)
 
-        blocks = []
-        while n > 0:
-            block = yield from self.read(n)
-            if not block:
-                partial = b''.join(blocks)
-                raise IncompleteReadError(partial, len(partial) + n)
-            blocks.append(block)
-            n -= len(block)
+            yield from self._wait_for_data('readexactly')
 
-        assert n == 0
-
-        return b''.join(blocks)
+        if len(self._buffer) == n:
+            data = bytes(self._buffer)
+            self._buffer.clear()
+        else:
+            data = bytes(self._buffer[:n])
+            del self._buffer[:n]
+        self._maybe_resume_transport()
+        return data
 
     if compat.PY35:
         @coroutine


socketpair:
I still not tested performance. Please be patient, or help me with that :)

asyncio/streams.py
```diff
@@ -658,25 +659,23 @@ def readexactly(self, n):
...
-            n -= len(block)
-
-        assert n == 0
-
-        return b''.join(blocks)
+        while True:
+            current_len = len(self._buffer)
+
+            if current_len >= n:
+                break
```
methane:
```py
while len(self._buffer) < n:
```

asyncio/streams.py
```diff
@@ -658,25 +659,23 @@ def readexactly(self, n):
...
+
+            if self._eof:
+                incomplete = self._buffer.copy()
```
methane:
```py
incomplete = bytes(self._buffer)
```

asyncio/streams.py
```diff
@@ -658,25 +659,23 @@ def readexactly(self, n):
...
+                self._buffer.clear()
+                raise IncompleteReadError(incomplete, n)
+
+            yield from self._wait_for_data('readexactly')
+
+        data = self._buffer[:n]
```
methane: If we should return exactly `bytes`,

```py
data = bytes(memoryview(self._buffer)[:n])
```

I wonder if bytearray has `.tobytes(offset=0, size=None)` method to create bytes directly...

asyncio/streams.py
socketpair: I will fix, but just curious, is it faster ?

socketpair: fixed

asyncio/streams.py
socketpair: fixed

asyncio/streams.py
socketpair: fixed

socketpair:
#394 should be fixed with that. Will benchmark after some time.

1st1:
@socketpair I can merge this PR in 3.6 today/tomorrow if you can show that it improves performance.

asyncio/streams.py
```diff
@@ -658,25 +659,18 @@ def readexactly(self, n):
...
+        while len(self._buffer) < n:
+            if self._eof:
+                incomplete = bytes(self._buffer)
+                self._buffer.clear()
+                raise IncompleteReadError(incomplete, n)
+
+            yield from self._wait_for_data('readexactly')
+
+        data = bytes(memoryview(self._buffer)[:n])
+        del self._buffer[:n]
```
socketpair: In non-cpython implementations, destructing of `memoryview` may be postponed, so cutting buffer may fail. In order to fix that, we should call `memoryview_obj.release()`...

Not sure if it is actual.

methane: Ah, you're right, I think.
I wonder if there is more simple and safe API.

socketpair: To overcome that annoying misfuture, It is require to make two separate variables: i.e.

```python
memview1 = memoryview(self._buffer)
memview2 = memview1[:n]
data = bytes(memview2)
memview2.release()
memview1.releas()
```

I decide to throw out that trash...Yes it should be faster, but....

gvanrossum: I think you should watch your language.

socketpair: Sorry, I have fixed comment.

socketpair:
I have benchmarked that.
8.72 on master branch
6.65 on this branch.

program:
```python
#!/usr/bin/env python3.5

import asyncio
import os
import time


def bench(fun):
    # todo: @wraps(fun)
    async def newfun(rs, chunksize):
        a = time.monotonic()
        await fun(rs, chunksize)
        b = time.monotonic()
        print('{}: chunk size={}: {:.2f} seconds.'.format(
            fun.__name__,
            chunksize,
            b - a,
        ))
    return newfun


@bench
async def action(rs, chunksize):
    try:
        while True:
            await rs.readexactly(chunksize)
    except asyncio.IncompleteReadError:
        pass

async def writer(ws, bufsize):
    ws.write(b'x' * bufsize)
    await ws.drain()
    ws.close()

async def amain():
    bufsize = 30 * 1024 * 1024
    chunksize = 16

    handler = lambda rs, ws: writer(ws, bufsize)
    server = await asyncio.start_server(handler, '127.0.0.1', 0)
    addr = server.sockets[0].getsockname()
    (rs, ws) = await asyncio.open_connection(*addr)
    server.close()
    await server.wait_closed()
    # now writing part is running "in background". TODO: wait it completion somehow
    await action(rs, chunksize)
    ws.close()


loop.run_until_complete(amain())
```

socketpair:
Adding memoryview slowers down (!) so, it is good not to use `memoryview` before copying in that place.

socketpair:
@gvanrossum @methane @1st1 

Any feedback ?

methane:
lgtm

asyncio/streams.py
```diff
@@ -658,25 +659,22 @@ def readexactly(self, n):
...
-        return b''.join(blocks)
+        while len(self._buffer) < n:
+            if self._eof:
+                incomplete = bytes(self._buffer)
+                self._buffer.clear()
+                raise IncompleteReadError(incomplete, n)
+
+            yield from self._wait_for_data('readexactly')
+
+        if len(self._buffer) == n:
```
methane: IMO, I want to add helper C function for "slicing bytes from bytearray".
And I hope it will be part of Python 3.7, (e.g. `bytes.frombuffer(buff, start=0, end=None)`).

socketpair:
@1st1 @gvanrossum ping

1st1:
I'll commit this later today.

1st1:
Committed by hand in b0e7438c2f7ca21135ddd89629f61185f699e17b. Thanks!

Issue #401, opened 2016-08-21
asyncio.Condition() in async with - dry


dbivolaru:
The following doesn't work. The equivalent in threading works.

    cond = asyncio.Condition()
    async with cond as c:
        await c.wait()

Considering the DRY principle, I think `asyncio/locks.py:_ContextManagerMixin.__aenter__` should return self instead of None:

        @coroutine
        def __aenter__(self):
            yield from self.acquire()
            return self

PEP492 actually doesn't mention anything about returning None. Iinstead it explicitly mentions `VAR = await aenter` in the pseudocode example [https://www.python.org/dev/peps/pep-0492/#new-syntax](url) and the example with lock is a few paragraphs afterwards.

Opinions? Did I miss anything?

asvetlov:
Your PR brings non-consistence with `threading` module.
```
with threading.Condition() as lock:
    ...
```
returns *lock status*, not `Condition` itself.

*lock status* is just `bool` value: 
* `True` for acquired lock
* `False` for try to lock but failure for timeout reason.

Since `asyncio` has no timeout for locking the return value from context manager should be always `True`.
Now it's `None`, which is not bad -- nobody will check *always true* conditions.

dbivolaru:
@asvetlov Thanks for the explanation. I should have looked more carefully in threading. It is still slightly against DRY and the way normal Contexts work. Will give the PEP's another read, if anything we should explicitly mention it in there if it's missing.

asvetlov:
Technically context manager may return any object, not itself.
It's rare but possible and makes sense in some contexts.
Anyway, changing semantic to not compatible with `threading` is worse than supporting existing approach (while the approach maybe is not very obvious).

From other hand you may think that acquiring lock from condition is an operation which returns something new (an information about acquirement) but not condition instance itself.

dbivolaru:
@gvanrossum I did not await a response but what I meant by DRY is that the last statement `#3` looks the cleanest.

    #1
    async with self.state[topic].conditions[id]:
        self.state[topic].conditions[id].wait()

    #2
    cond = self.state[topic].conditions[id]
    async with cond:
        cond.wait()

    #3
    async with self.state[topic].conditions[id] as c:
        c.wait()

[...]

PR #402, opened 2016-08-21
Bugfix for __aenter__ in asyncio/locks.py

dbivolaru:
See #401


asvetlov:
See explanation for #401

PR #408, opened 2016-08-31
Fix ordering issues in UNIX read/write pipe transport constructors

ronf:
Here's a pull request designed to fix the issue reported in #368.

This commit re-orders the initialization code in the _UnixReadPipeTransport
and _UnixWritePipeTransport constructors to make sure all members are
assigned a value, even in the case where ValueError is raised due to
an incompatible type of pipe. This avoids exceptions being raised in
__repr__() due to the missing members.

In the case where ValueError is raised, this commit also clears the values
of _pipe, _fileno, and _protocol since this transport isn't returned,
avoiding a spurious "unclosed transport" warning when the object is
garbage-collected.

### DIFF
--- a/asyncio/unix_events.py
+++ b/asyncio/unix_events.py
@@ -305,14 +305,20 @@ class _UnixReadPipeTransport(transports.ReadTransport):
         self._loop = loop
         self._pipe = pipe
         self._fileno = pipe.fileno()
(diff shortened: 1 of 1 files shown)


asvetlov:
Thank you, @ronf

ronf:
Thank you!

PR #410, opened 2016-09-01
Fix an unclosed transport error in test_events

ronf:
When running the asyncio unit tests, I noticed that I was seeing the
following error from test_ssl_connect_accepted_socket():

/Volumes/ronf/src/asyncio/asyncio/sslproto.py:328: ResourceWarning: unclosed transport <asyncio.sslproto._SSLProtocolTransport object at 0x10d7d57b8>
  warnings.warn("unclosed transport %r" % self, ResourceWarning)

This commit fixes this error by calling close on the transport which is
opened rather than calling close on the socket of the accepted
connection used to create it.


asvetlov:
thanks

Issue #367, opened 2016-07-01
UDP equivalent of asyncio.streams.open_connection


fr0stbite:
Hello,

I was experimenting with `asyncio` a lot lately and I noticed that it provides very small support for writing asynchronnous UDP applications. I wanted to write a very simple module, that would allow me to open a UDP socket, send some data over it and close it immediately. I tried to accomplish this by implementing my own `DatagramProtocol`.

```python
class SendProtocol:

    def connection_made(self, transport):
        transport.sendto(data)
        transport.close()
```

However, this implementation produces an error everytime the `transport.close` method is called, because the `_read_ready` action fails.

I was trying to find something that would help me in the docs, and I came across [`asyncio.streams.open_connection`](https://docs.python.org/3/library/asyncio-stream.html#asyncio.open_connection) method, that returns `StreamReader` and `StreamWriter` objects, which I could use to control the flow of my application. Unfortunately, this method is only compatible with TCP, not UDP. I would like to ask if there's an equivalent of this method for UDP transport that I overlooked. Or if you plan to add it in the future.

Cheers!

gvanrossum:
Can you show the whole (minimal) program, and the specific error you got?
Do you think the error is due to calling transport.close() inside your
connection_made()?

A streams wrapper around UDP would set entirely wrong expectations, since
streams assume reliable byte streams where the boundaries between packages
are irrelevant (like TCP), but UDP provides lossy datagrams.

fr0stbite:
I've written some simple examples to demonstrate the issue. You can read the source code below.

```python
# client.py

import asyncio

class SendProtocol:
    def __init__(self, message, loop):
        self.message = message
        self.loop = loop

    def connection_made(self, transport):
        print("Sending:", self.message)
        transport.sendto(self.message.encode())
        transport.close()

    def datagram_received(self, data, addr):
        print("Received:", data.decode())

    def error_received(self, exception):
        print("Error occurred:", exception)

    def connection_lost(self, exception):
        print("Socket closed.")
        self.loop.stop()

loop = asyncio.get_event_loop()
address = ("localhost", 9999)
message = "Hello World!"

connect = loop.create_datagram_endpoint(
    lambda: SendProtocol(message, loop),
    remote_addr=address
)
transport, protocol = loop.run_until_complete(connect)

loop.run_forever()
loop.close()

```

```python
# server.py

import asyncio

class DummyServerProtocol:

    def connection_made(self, transport):
        pass

    def datagram_received(self, data, address):
[...]

PR #419, opened 2016-09-16
Improve handling of stale fds for SelectSelector

SethMichaelLarson:
Made SelectSelector more resilient to stale file descriptors which would result in strange outcomes such as in Issue #367. Removed using `select.select` directly as non-Windows SelectSelector (should?) break in the same situation as Issue #367 if SelectSelector were the default on a non-Windows platform as well. Added a new test for a stale file descriptor for all SelectSelectors.


gvanrossum:
Meta-comment: the selectors module is not a part of asyncio -- it's in the stdlib (we just bundled it for the convenience of Python 3.3 users).

SethMichaelLarson:
@gvanrossum Oh, well that makes things interesting. Is there perhaps a way to allow the use of our own SelectSelector if it's on Windows? Because as it stands that select cannot handle stale fds.

SethMichaelLarson:
Fixed a quick bug where if all the watched fds were stale Windows would lock up. Hopefully that'll pass on AppVeyor.

1st1:
> @gvanrossum Oh, well that makes things interesting. Is there perhaps a way to allow the use of our own SelectSelector if it's on Windows? Because as it stands that select cannot handle stale fds.

What Guido meant was that you should file this bug to http://bugs.python.org since `selectors` module isn't part of asyncio (it's only included in this repo for compatibility with Python 3.3). That's how the process is defined for now.

SethMichaelLarson:
@1st1 Shame, git/GitHub is hard enough to use on Windows. I don't really want to get Mercurial just for one small patch just to have CPython switch back to GitHub. I guess close this PR.

1st1:
> @1st1 Shame, git/GitHub is hard enough to use on Windows. I don't really want to get Mercurial just for one small patch just to have CPython switch back to GitHub. I guess close this PR.

Well, CPython repo will be migrated to GitHub in a couple of months. If this PR fixes a bug, please at least open an issue and attach the patch on bugs.python.org. I don't have capacity to review your PR in the next couple of weeks :(

SethMichaelLarson:
@1st1 Does fix a bug, I'll open an issue there.

1st1:
@SethMichaelLarson so did you open an issue on bugs.python.org for this?

PR #420, opened 2016-09-16
Raise RuntimeError when transport's FD is used with add_reader etc

### LINKED ISSUE #372: Using transport's socket with low level add_reader/add_writer
Turns out people use transports' sockets with `add_reader` / `add_writer` low level APIs.  I've discovered this while debugging a crash of a webapp deployed with uvloop. The reason of that crash is how aiohttp implements sendfile: they use `transport.get_extra_info('socket')` to get the underlying socket, and then they use `loop.add_writer` and `loop.remove_writer` on that socket.

The crash in uvloop is caused by how file descriptors are stored internally by libuv.  I'll make a workaround for that by using `os.dup()` to return a duplicate socket from `transport.get_extra_info('socket')`.

However, I think we should do this in asyncio too. The thing is that when you use `add_writer` and `remove_writer` on the transport's socket, you're messing with the internal state of the transport. For instance, a transport might be in the middle of writing data, and calling `remove_writer` will cause the whole program to hang.

I see two options:

1. We always return duplicate sockets from `transport.get_extra_info()`, hiding the actual socket that transport is attached to.

2. We find a way to raise an exception if `add|remove_witer|reader` is used on a transport's socket.

I'm more inclined to do the 1.

/cc @asvetlov

1st1:
This PR fixes #372.

The idea is that low-level socket APIs should refuse to work with FDs that belong to some `Transport`.

This PR adds private implementations for `add_writer`, `add_reader`, `remove_writer`, and `remove_reader`. Transports **only** call private implementation.

Public implementations of those methods do an extra check on the FD, raising a `RuntimeError` when the FD belongs to some `Transport`. `loop.sock_sendall`, `loop.sock_connect`, `loop.sock_accept` and `loop.sock_recv` only call public methods.

FWIW uvloop has a lot of functional network unittests, all of them pass on asyncio with this patch.

### DIFF
--- a/asyncio/selector_events.py
+++ b/asyncio/selector_events.py
@@ -252,9 +252,10 @@ class BaseSelectorEventLoop(base_events.BaseEventLoop):
         except KeyError:
             pass
         else:
-            raise RuntimeError(
-                'File descriptor {!r} is used by transport {!r}'.format(
-                    fd, transport))
+            if not transport.is_closing():
+                raise RuntimeError(
+                    'File descriptor {!r} is used by transport {!r}'.format(
+                        fd, transport))
 
     def _add_reader(self, fd, callback, *args):
         self._check_closed()
--- a/asyncio/selector_events.py
+++ b/asyncio/selector_events.py
@@ -11,6 +11,7 @@ import errno
 import functools
 import socket
 import warnings
+import weakref
 try:
     import ssl
 except ImportError:  # pragma: no cover
@@ -64,6 +65,7 @@ class BaseSelectorEventLoop(base_events.BaseEventLoop):
         logger.debug('Using selector: %s', selector.__class__.__name__)
         self._selector = selector
         self._make_self_pipe()
+        self._transports = weakref.WeakValueDictionary()
 
     def _make_socket_transport(self, sock, protocol, waiter=None, *,
                                extra=None, server=None):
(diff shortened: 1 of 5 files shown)


asvetlov:
After failed tests fixing, of course.

1st1:
Thank you Andrew!  @gvanrossum would be cool if you can take a quick look over this PR, I want to merge it before beta2.

gvanrossum:
I'm confused. Where is `self._transports` ever updated? It seems it's always empty, so the check will never fail. What am I missing?

1st1:
@gvanrossum Transports register when they are being constructed: https://github.com/python/asyncio/pull/420/files#diff-ddadd6cf041d620407e07c7c4f1b0149R572

gvanrossum:
I see. What about transports that don't inherit from _SelectorTransport? I guess those don't get this error when misused in debug mode but otherwise no harm is done right?

If that's so I think this is fine for b2.

asvetlov:
Random thought.
Let's assume I want to make my own transport for some reason.
Honestly I did it only in `aiozmq` (ZMQ sockets are very different from regular ones, you know) but maybe somebody will need it for TCP sockets for some reason.
Who knows, in aiohttp we have own StreamReader for working with HTTP chunks etc.

It would be cool to have `loop` methods for *fd protecting*  by protect/unprotect calls.
Not in 3.6, sure -- but maybe in 3.7?

1st1:
@gvanrossum I think we cover all transports except `SubprocessTransport`, but those don't expose the socket via `get_extra_info`.  BTW, the check is always enabled, even in production mode (keeping a `WeakValueDict` for transports is cheap).

> Honestly I did it only in aiozmq (ZMQ sockets are very different from regular ones, you know) only but maybe somebody will need it for TCP sockets for some reason.

@asvetlov Let's think about that after 3.6 :)

asvetlov:
> @asvetlov Let's think about that after 3.6 :)
Agree

1st1:
Committed by hand, closing this PR now.

Issue #429, opened 2016-09-28
CoroWrapper in debug mode has incorrect throw method signature


Bahus:
Hello!

 When `PYTHONASYNCIODEBUG` is set asyncio wraps each coroutine in special `CoroWrapper`class which has [incorrect signature](https://github.com/python/asyncio/blob/b100957e7baf9482509631e504160d6c59509d26/asyncio/coroutines.py#L123) for `throw` method:

```python
def throw(self, exc):
    return self.gen.throw(exc)
```

However it should be:

```python
def throw(self, type, value=None, traceback=None):
    return self.gen.throw(type, value, traceback)
```

For example the problem appears when I'm using Tornado web-server with enabled asyncio debugging mode exception thrown in a coroutine causes another one:

```
Traceback (most recent call last):
  File ".../lib/python3.5/site-packages/tornado/web.py", line 1445, in _execute
    result = yield result
  File ".../lib/python3.5/site-packages/tornado/gen.py", line 1008, in run
    value = future.result()
  File ".../lib/python3.5/site-packages/tornado/concurrent.py", line 232, in result
    raise_exc_info(self._exc_info)
  File "<string>", line 3, in raise_exc_info
  File ".../lib/python3.5/site-packages/tornado/gen.py", line 1014, in run
    yielded = self.gen.throw(*exc_info)
  File "<string>", line 6, in _wrap_awaitable
TypeError: throw() takes 2 positional arguments but 4 were given
```

See tornado's code [here](https://github.com/tornadoweb/tornado/blob/24a8b2256c4da30823c4017f31356ce67d9ba389/tornado/gen.py#L1040).

Martiusweb:
Hi,

Indeed, it seems that CoroWrapper.throw should have the same signature than generator.throw():
https://docs.python.org/3/reference/expressions.html#generator.throw
However, the documentation doesn't say that it is possible to pass an exception instance if type is the only argument.

gvanrossum:
Should be an easy fix. Just send a PR.

--Guido (mobile)

soar:
What about fix for this problem?

1st1:
The change to fix this was committed a year ago, so Python 3.6 should have it fixed.

soar:
I use tornado like this:

```python
tornado.platform.asyncio.AsyncIOMainLoop().install()

loop = asyncio.get_event_loop()
loop.set_debug(True)

tornado_app = app.webapp.TornadoApp()
tornado_app.listen(8080)

[...]

PR #430, opened 2016-09-28
make CoroWrapper.throw fully compatible with gen.throw

Martiusweb:
Following #429, this PR makes the signature of CoroWrapper.throw compatible with gen.throw.

The signature of gen.throw is the same in python 3.3 - 3.5. Unless I'm missing something, the patch shouldn't change anything for the current implementation of asyncio, since throw is only called explicitly in Task:
https://github.com/python/asyncio/blob/master/asyncio/tasks.py#L241

### DIFF
--- a/asyncio/coroutines.py
+++ b/asyncio/coroutines.py
@@ -120,8 +120,8 @@ class CoroWrapper:
         def send(self, value):
             return self.gen.send(value)
 
-    def throw(self, exc):
-        return self.gen.throw(exc)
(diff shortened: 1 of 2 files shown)


gvanrossum:
Since the `CoroWrapper` class tries to emulate a regular coroutine/generator object I agree that this should be considered a bugfix. Can you add a unittest?

gvanrossum:
Thanks! Looking forward to more from you.

PR #431, opened 2016-09-28
Trim off finished futures from the Event._waiters

ZhukovAlexander:
Current implementation of Event.wait uses dequeue.remove(e),
wich has O(n) complexity. We can do better by utilizing this
time to also remove all finished futures from the queue
on the way up until we reach the future that is in a pending
state. This will make subsequent calls to Event.wait or
Event.set to perform less work on a waiters queue

### DIFF
--- a/asyncio/locks.py
+++ b/asyncio/locks.py
@@ -264,8 +264,12 @@ class Event:
             yield from fut
             return True
         finally:
-            self._waiters.remove(fut)
-
+            # popleft all finished futures up to the first pending one
+            while self._waiters:
+                if not self._waiters[0].done():
+                    break
+                else:
+                    self._waiters.popleft()
 
 class Condition(_ContextManagerMixin):
     """Asynchronous equivalent to threading.Condition.


asyncio/locks.py
```diff
@@ -264,8 +264,12 @@ def wait(self):
...
             return True
         finally:
-            self._waiters.remove(fut)
-
+            # popleft all finished futures up to the first pending one
+            while self._waiters:
+                if not self._waiters[0].done():
+                    break
+                else:
+                    self._waiters.popleft()
```
gvanrossum: I'd like to see some kind of proof/assertion that once we fall out of the loop `fut` has in fact been removed.

SethMichaelLarson:
Pretty sure that `fut` is always at index 0 within `_waiters` by the time that it yields and `remove()` only removes the first encountered element so it'll stop there. Instead of using `remove()` could just use `popleft()` as it's probably faster.

ZhukovAlexander:
@SethMichaelLarson , I think you are right. Though, the side effect of this algorithm is that it will also trim all finished futures that come after the current one, so it still has an effect.

SethMichaelLarson:
@ZhukovAlexander When would that occur? Each future trims itself as it's finished so there's only ever one finished future in `_waiters`. The only time I can think where that would happen is if code was muddling with the internals.

asyncio/locks.py
ZhukovAlexander: Would checking if the poped future is the one we care about and raising a RuntimeError if it wasn't found work?
````python
else:
    found = self._waiters.popleft() is fut
if not found:
    raise RuntimeError
````

gvanrossum:
If fut always occurs at index 0 there is no O() improvement from using popleft(), so the point of the diff is moot -- let's close this PR then.

PR #433, opened 2016-09-29
Document gather's undeterministic scheduling of awaitables and coroutines

### LINKED ISSUE #432: gather unexpectedly schedules coroutines in undeterministic order
If you pass awaitables or coroutines to `asyncio.gather`, the function takes responsibility for wrapping them in futures and scheduling their execution. The order in which they are scheduled for execution is non-deterministic due to `gather` coercing the arguments to a unique set to ensure the same coroutine object isn't undertaken more than once.

Example:
```python
import asyncio

order = []
@asyncio.coroutine
def append_a_thing(thing):
    order.append(thing)
    return thing

task = asyncio.gather(append_a_thing(1), append_a_thing(2), append_a_thing(3), append_a_thing(4))
loop.run_until_complete(results)

print("Things appended were {}.".format(task.result()))
print("Order of execution was {}.".format(order))
```
This might print:
```
Things appended were [1, 2, 3, 4].
Order of execution was [3, 1, 2, 4].
```

Either this is cool, and we should just warn the programmer that their coroutines may be scheduled (and executed) in any order their Python interpreter chooses…
or it's not cool and we should schedule tasks in the order expected by the programmer, while maintaining that unique awaitables and coroutines are only scheduled for execution once.

I'll submit a pull request for either scenario and someone else can decide what's cool.

JustinTArthur:
Resolves #432. This is probably the best performing solution. The workaround for a programmer that wishes their coroutine objects to be undertaken in a deterministic order is for them to schedule the execution on the loop themselves with `asyncio.ensure_future` or `loop.create_task` before calling `asyncio.gather`.

### DIFF
--- a/asyncio/tasks.py
+++ b/asyncio/tasks.py
@@ -594,6 +594,10 @@ def gather(*coros_or_futures, loop=None, return_exceptions=False):
     """Return a future aggregating results from the given coroutines
     or futures.
 
+    Coroutines will be wrapped in a future and scheduled in the event
+    loop. They will not necessarily be scheduled in the same order as
+    passed in.
+
     All futures must share the same event loop.  If all the tasks are
     done successfully, the returned future's result is the list of
     results (in the order of the original sequence, not necessarily


JustinTArthur:
#434 is an alternative to this.

PR #434, opened 2016-09-29
Ensure order of tasks scheduled on event loop by gather is deterministic.

### LINKED ISSUE #432: gather unexpectedly schedules coroutines in undeterministic order

JustinTArthur:
Resolves #432. Likely a wee bit slower in most Python implementations as before this change, the interpreter had no obligation to order when uniqueness was established via a simple construction of a `set`. However, the programmer can now achieve a predictable order of scheduling (and possibly execution) of their awaitables or coroutines.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -16,6 +16,7 @@ Guido van Rossum <guido AT python.org>: creator of the asyncio project and autho
 Gustavo Carneiro <gjcarneiro AT gmail.com>
 Jeff Quast
 Jonathan Slenders <jonathan.slenders AT gmail.com>
+Justin Turner Arthur <justinarthur AT gmail.com>
 Nikolay Kim <fafhrd91 AT gmail.com>
 Richard Oudkerk <shibturn AT gmail.com>
 Saúl Ibarra Corretgé <saghul AT gmail.com>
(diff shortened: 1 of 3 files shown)


JustinTArthur:
#433 is an alternative to this.

PR #436, opened 2016-09-29
Use staticmethod or module function when instance isn't referenced.

JustinTArthur:
I used `staticmethod`s where I thought the taxonomy or overridability might be useful, or I didn't know.

### DIFF
--- a/asyncio/base_events.py
+++ b/asyncio/base_events.py
@@ -76,6 +76,20 @@ def _format_pipe(fd):
         return repr(fd)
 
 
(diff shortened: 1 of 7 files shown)


Martiusweb:
Hi,

Do you have a use case where overriding those functions is useful?
Imho, since those functions and methods are private, and unless there is a good use case inside the codebase of asyncio, we shouldn't encourage someone to override them.

gvanrossum:
This looks like a refactoring that accomplishes nothing except code churn. Let's not do this.

PR #437, opened 2016-09-30
Don't use getservbyname for non-numeric ports.

### LINKED ISSUE #422: socket.getservbyname() in _ipaddr_info could block
#354 added a call to socket.getservbyname() in _ipaddr_info() to resolve service names into port numbers.  However, local system configuration could result in getservbyname() blocking for some time, especially if the system is configured to query something like NIS or use the old hesiod protocol to look up this information.  Even in the default configuration, if the disk holding /etc/services is unhappy with life, attempting to read /etc/services could take some time.  This would bring the whole event loop to a grinding halt.  getaddrinfo() (usually) will handle non-numeric port strings so long as AI_NUMERICSERV is not used, so it would probably be best to just let getaddrinfo() handle it.

(While looking into this, I discovered the Python getservbyname() implementation is not thread-safe, but that's a different bug...)

ajdavis:
Fixes #422.

### DIFF
--- a/asyncio/base_events.py
+++ b/asyncio/base_events.py
@@ -115,24 +115,16 @@ def _ipaddr_info(host, port, family, type, proto):
 
     if port is None:
         port = 0
-    elif isinstance(port, bytes):
-        if port == b'':
-            port = 0
-        else:
-            try:
-                port = int(port)
-            except ValueError:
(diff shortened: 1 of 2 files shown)


gvanrossum:
Thanks!

PR #440, opened 2016-10-05
starttls: Add "call_connection_made" arg to SSLProtocol.__init__

1st1:
This is the key modification of `sslproto.py` required to implement starttls. With this patch in place, we can implement starttls as an external package without copying/pasting of `sslproto.py`.

The patch is super small and non-invasive, only adding an extra argument to `SSLProtocol.__init__` (everything inside `sslproto.py` is an implementation detail of asyncio).

### DIFF
--- a/asyncio/sslproto.py
+++ b/asyncio/sslproto.py
@@ -410,7 +410,8 @@ class SSLProtocol(protocols.Protocol):
     """
 
     def __init__(self, loop, app_protocol, sslcontext, waiter,
-                 server_side=False, server_hostname=None):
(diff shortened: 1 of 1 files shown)


1st1:
@gvanrossum Are you OK with me merging this in 3.6 beta 2?

gvanrossum:
Yes, LGTM.

1st1:
Thanks! Merged by hand in d4abdd505ff609b7a8799bc405bfe476898b20ae. I'll make a PyPI package for starttls soon. Hopefully, this PR is enough (my current understanding is that it should be).

Issue #425, opened 2016-09-22
unlink stale unix socket before binding


socketpair:
`loop.create_unix_server()` should remove socket at that path if it exists. Every application that wants to use unix socket server should do that. Why not to add code that removes unix socket before binding ?

It should:
1. Check if it is exists and that is UNIX socket (i.e. do `os.stat()`)
2. raise exception if removal failed, raise exception if this path exists but is not unix socket.
3. not even try to remove if path starts with zero byte (i.e. abstract unix socket)

gjcarneiro:
What if you are trying, by mistake, to start a second instance of a server, with the previous one still running?  We want to catch that mistake, instead of silently removing the existing socket.

socketpair:
All software that I see unlinks stale socket unconditionally. Can you show any one that checks for the other instance ? Also, seems the kernel checks if address already in use during `bind()` even when FS node is removed.

1st1:
+1 one to do this. @socketpair can you quickly make a PR (so that we can fix this before 3.6b2)?

1st1:
Although there is another approach: maybe the correct thing to do is to remove the socket file after we close the unix server? How Twisted/Tornado implement this?

ZhukovAlexander:
@1st1 both Twisted and Tornado clean stale sockets before binding:

- https://github.com/twisted/twisted/blob/trunk/src/twisted/internet/unix.py#L290
- https://github.com/tornadoweb/tornado/blob/master/tornado/netutil.py#L215

socketpair:
@1st1 @ZhukovAlexander @gjcarneiro please review

PR #441, opened 2016-10-06
#425 Remove UNIX socket from FS before binding.

socketpair:


asyncio/unix_events.py
```diff
@@ -253,6 +253,15 @@ def create_unix_server(self, protocol_factory, path=None, *,
 
             sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
 
+            # Check for abstract socket. `str` and `bytes` paths are supported.
+            if path[0] not in (0, '\x00'):
+                try:
+                    if stat.S_ISSOCK(os.stat(path).st_mode):
+                        os.remove(path)
```
ZhukovAlexander: What if the file exists and is not a socket?

socketpair: `if stat.S_ISSOCK(os.stat(path).st_mode)`

As much s we can. Let's kernel to decide what to do during `bind()`

1st1: I don't understand: can `stat.S_ISSOCK` return `True` if it's not a socket?

1st1:
I think we should also remove the socket *after* the server is closed.

asyncio/unix_events.py
socketpair: No, it can not. If this path refers to plain file, `bind syscall will fail`. We must not erase non-sockets in any case.

asyncio/unix_events.py
```diff
@@ -253,6 +253,15 @@ def create_unix_server(self, protocol_factory, path=None, *,
...
+                except OSError:
+                    # Directory may have permissions only to create socket.
+                    pass
```
socketpair: Sure, but we should introduce something like `log = logging.getLogger(__name__)`. Please tell where to do that.

tests/test_unix_events.py
```diff
@@ -243,9 +243,7 @@ def test_create_unix_server_existing_path_sock(self):
             sock.bind(path)
             with sock:
                 coro = self.loop.create_unix_server(lambda: None, path)
-                with self.assertRaisesRegex(OSError,
-                                            'Address.*is already in use'):
-                    self.loop.run_until_complete(coro)
+                self.loop.run_until_complete(coro)
```
socketpair: Seems, potential client will connect to second socket. First one will be hidden.

socketpair: I did not check. Maybe it is implementation specific - i.e. Different on bsd and linux.

asyncio/unix_events.py
1st1: Look at how `base_events.py` logs stuff.

asyncio/unix_events.py
1st1: Ah, OK, I now understand what was @ZhukovAlexander's question about.

tests/test_unix_events.py
1st1: Please try to figure out what exactly is going on in this unittest.

1st1:
@socketpair FWIW we're just a couple days from a complete feature freeze of asyncio. Will you have time to finish this, or should I submit a new PR?

asyncio/unix_events.py
```diff
@@ -272,9 +281,10 @@ def create_unix_server(self, protocol_factory, path=None, *,
                 raise ValueError(
                     'path was not specified, and no sock specified')
 
-            if sock.family != socket.AF_UNIX:
+            if sock.family != socket.AF_UNIX or sock.type != socket.SOCK_STREAM:
```
socketpair: done

socketpair:
> I think we should also remove the socket after the server is closed.
@1st1 It is not so easy, to do that, I have to intorduce something new in `base_events.Server`

Actually (as I think), removing before binding and removing after closing should be done at Python's `socket` level, not at asyncio level. But this is another sad story.

1st1:
>> I think we should also remove the socket after the server is closed.
@1st1 It is not so easy, to do that, I have to intorduce something new in base_events.Server

> Actually (as I think), removing before binding and removing after closing should be done at Python's socket level, not at asyncio level. But this is another sad story.

Maybe. But let's fix this in asyncio first.

asyncio/unix_events.py
socketpair: fixed

tests/test_unix_events.py
socketpair: I have checked. If socket-inode on FS is deleted, no one can connect. If it overrided with another UNIX-socket, all clients will be connected to that new one, as I expect eariler.

```python
import socket
import os
if os.path.exists('/tmp/qwe'):
    os.unlink('/tmp/qwe')
sock1 = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
sock1.bind('/tmp/qwe')
sock1.listen()
os.unlink('/tmp/qwe')

sock2 = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
sock2.bind('/tmp/qwe')
sock2.listen()
#os.unlink('/tmp/qwe')

sock3 = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
sock3.connect('/tmp/qwe')
sock3.send(b'asd')

(sk, addr) = sock2.accept()
print(sk.recv(6))
```

socketpair:
> Maybe. But let's fix this in asyncio first.
@1st1 Can you advice best way to add removing after closing ?

tests/test_unix_events.py
1st1: I didn't know this, we need a comment somewhere (maybe in create_unix_server?)  Don't think we need this unittest then.

1st1: BTW, can you point out where you found this? Is this specified in POSIX?

1st1:
> @1st1 Can you advice best way to add removing after closing ?

After giving this some thought - let's not add this. I now remember an issue with uvloop: originally I was removing the path, but then someone's program broke because it was cleaning up the path manually and assumed it's always there.

Trying to remove the path before connecting to it makes total sense, removing it after is nice, but not strictly necessary.

asyncio/unix_events.py
socketpair: Do you still have a questions ?

1st1: No, no further questions here about this line :)

tests/test_unix_events.py
socketpair: I think, that unittests are fine. Unittests must check that all assertions are not failed on target OS.

1st1: What assertions?  Clearly, there is some system-depending behaviour here: some systems allow Unix sockets to bind to existing in-use Unix sockets paths.  Why should we have a test for that?

socketpair: Well. I will remove that test.

socketpair: I transform that test in a way it check that binding to same path as CLOSED socket works. This will improve code-coverage. And also I have found a bug with logging :).

1st1:
Thank you, Mark!

PR #444, opened 2016-10-07
Prohibit Tasks to await on themselves.

1st1:
This PR fixes http://bugs.python.org/issue27972

/cc @gvanrossum


1st1:
Committed by hand in https://github.com/python/asyncio/commit/e3fed68754002000be665ad1a379a747ad9247b6.

PR #452, opened 2016-11-03
Make get_event_loop() return the current loop if called from coroutines

1st1:
As discussed in [1] and [2], this PR makes `asyncio.get_event_loop()` to always return the currently running event loop when it is called from a coroutine.

The patch adds two new functions: `asyncio._set_current_loop()` and `asyncio._get_current_loop()`.  Both functions are thread-specific and are considered as low-level APIs, intended to be used by third-party loop implementations only. Even though they have a leading underscore, they will be documented as public asyncio API.

`loop.run_forever()` is modified to call `asyncio._set_current_loop()` and `asyncio._get_current_loop()`.

`asyncio.get_event_loop()` is modified to check if there is a current loop set by calling `asyncio._get_current_loop()`.  If there is no current event loop, it returns the result of `get_event_loop_policy().get_event_loop()` (old behaviour).

The biggest part of the patch is fixing unit tests: all test classes call `super().setUp()`. `test_utils.TestCase.setUp` patches `asyncio._get_current_loop()` to always return `None`. This change is to make sure that the loop is still being passed explicitly within asyncio.

[1] PR: https://github.com/python/asyncio/pull/355
[2] https://groups.google.com/d/msg/python-tulip/yF9C-rFpiKk/tk5oA3GLHAAJ

### DIFF
--- a/asyncio/base_events.py
+++ b/asyncio/base_events.py
@@ -393,7 +393,10 @@ class BaseEventLoop(events.AbstractEventLoop):
         """Run until stop() is called."""
         self._check_closed()
         if self.is_running():
-            raise RuntimeError('Event loop is running.')
+            raise RuntimeError('This event loop is already running')
+        if events._get_running_loop() is not None:
+            raise RuntimeError(
+                'Cannot run the event loop while another loop is running')
         self._set_coroutine_wrapper(self._debug)
         self._thread_id = threading.get_ident()
         if self._asyncgens is not None:
@@ -401,6 +404,7 @@ class BaseEventLoop(events.AbstractEventLoop):
             sys.set_asyncgen_hooks(firstiter=self._asyncgen_firstiter_hook,
                                    finalizer=self._asyncgen_finalizer_hook)
         try:
+            events._set_running_loop(self)
             while True:
                 self._run_once()
                 if self._stopping:
@@ -408,6 +412,7 @@ class BaseEventLoop(events.AbstractEventLoop):
         finally:
             self._stopping = False
             self._thread_id = None
+            events._set_running_loop(None)
             self._set_coroutine_wrapper(False)
             if self._asyncgens is not None:
                 sys.set_asyncgen_hooks(*old_agen_hooks)
--- a/asyncio/events.py
+++ b/asyncio/events.py
@@ -607,6 +607,30 @@ _event_loop_policy = None
 _lock = threading.Lock()
 
 
+# A TLS for the running event loop, used by _get_running_loop.
+class _RunningLoop(threading.local):
+    _loop = None
+_running_loop = _RunningLoop()
+
+
+def _get_running_loop():
+    """Return the running event loop or None.
+
+    This is a low-level function intended to be used by event loops.
+    This function is thread-specific.
+    """
+    return _running_loop._loop
+
+
+def _set_running_loop(loop):
+    """Set the running event loop.
+
+    This is a low-level function intended to be used by event loops.
+    This function is thread-specific.
+    """
+    _running_loop._loop = loop
+
+
 def _init_event_loop_policy():
     global _event_loop_policy
     with _lock:
@@ -632,7 +656,17 @@ def set_event_loop_policy(policy):
 
 
 def get_event_loop():
-    """Equivalent to calling get_event_loop_policy().get_event_loop()."""
+    """Return an asyncio event loop.
+
+    When called from a coroutine or a callback (e.g. scheduled with call_soon
+    or similar API), this function will always return the running event loop.
+
+    If there is no running event loop set, the function will return
+    the result of `get_event_loop_policy().get_event_loop()` call.
+    """
+    current_loop = _get_running_loop()
+    if current_loop is not None:
+        return current_loop
     return get_event_loop_policy().get_event_loop()
 
 
--- a/asyncio/test_utils.py
+++ b/asyncio/test_utils.py
@@ -449,7 +449,13 @@ class TestCase(unittest.TestCase):
         self.set_event_loop(loop)
         return loop
 
+    def setUp(self):
+        self._get_running_loop = events._get_running_loop
+        events._get_running_loop = lambda: None
+
     def tearDown(self):
+        events._get_running_loop = self._get_running_loop
+
         events.set_event_loop(None)
 
         # Detect CPython bug #23353: ensure that yield/yield-from is not used
--- a/tests/test_events.py
+++ b/tests/test_events.py
@@ -2233,6 +2233,7 @@ def noop(*args, **kwargs):
 class HandleTests(test_utils.TestCase):
 
     def setUp(self):
+        super().setUp()
         self.loop = mock.Mock()
         self.loop.get_debug.return_value = True
 
@@ -2411,6 +2412,7 @@ class HandleTests(test_utils.TestCase):
 class TimerTests(unittest.TestCase):
 
     def setUp(self):
+        super().setUp()
         self.loop = mock.Mock()
 
     def test_hash(self):
@@ -2718,6 +2720,27 @@ class PolicyTests(unittest.TestCase):
         asyncio.set_event_loop_policy(policy)
         self.assertIs(policy, asyncio.get_event_loop_policy())
         self.assertIsNot(policy, old_policy)
+
+    def test_get_event_loop_returns_running_loop(self):
+        class Policy(asyncio.DefaultEventLoopPolicy):
+            def get_event_loop(self):
+                raise NotImplementedError
+
+        loop = None
+
+        old_policy = asyncio.get_event_loop_policy()
+        try:
+            asyncio.set_event_loop_policy(Policy())
+            loop = asyncio.new_event_loop()
+
+            async def func():
+                self.assertIs(asyncio.get_event_loop(), loop)
+
+            loop.run_until_complete(func())
+        finally:
+            asyncio.set_event_loop_policy(old_policy)
+            if loop is not None:
+                loop.close()
 
 
 if __name__ == '__main__':
--- a/tests/test_base_events.py
+++ b/tests/test_base_events.py
@@ -154,6 +154,7 @@ class BaseEventTests(test_utils.TestCase):
 class BaseEventLoopTests(test_utils.TestCase):
 
     def setUp(self):
+        super().setUp()
         self.loop = base_events.BaseEventLoop()
         self.loop._selector = mock.Mock()
         self.loop._selector.select.return_value = ()
@@ -976,6 +977,7 @@ class MyDatagramProto(asyncio.DatagramProtocol):
 class BaseEventLoopWithSelectorTests(test_utils.TestCase):
 
     def setUp(self):
+        super().setUp()
         self.loop = asyncio.new_event_loop()
         self.set_event_loop(self.loop)
 
@@ -1690,6 +1692,24 @@ class BaseEventLoopWithSelectorTests(test_utils.TestCase):
         self.assertRegex(fmt % tuple(args),
                          "^Executing <Task.*stop_loop_coro.*> "
                          "took .* seconds$")
+
+
+class RunningLoopTests(unittest.TestCase):
+
+    def test_running_loop_within_a_loop(self):
+        @asyncio.coroutine
+        def runner(loop):
+            loop.run_forever()
+
+        loop = asyncio.new_event_loop()
+        outer_loop = asyncio.new_event_loop()
+        try:
+            with self.assertRaisesRegex(RuntimeError,
+                                        'while another loop is running'):
+                outer_loop.run_until_complete(runner(loop))
+        finally:
+            loop.close()
+            outer_loop.close()
 
 
 if __name__ == '__main__':
--- a/tests/test_futures.py
+++ b/tests/test_futures.py
@@ -79,6 +79,7 @@ class DuckFuture:
 class DuckTests(test_utils.TestCase):
 
     def setUp(self):
+        super().setUp()
         self.loop = self.new_test_loop()
         self.addCleanup(self.loop.close)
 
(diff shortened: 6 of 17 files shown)


asvetlov:
Looks good!

asyncio/base_events.py
```diff
@@ -400,6 +400,8 @@ def run_forever(self):
             old_agen_hooks = sys.get_asyncgen_hooks()
             sys.set_asyncgen_hooks(firstiter=self._asyncgen_firstiter_hook,
                                    finalizer=self._asyncgen_finalizer_hook)
+        old_loop = events._get_current_loop()
+        events._set_current_loop(self)
```
gvanrossum: This should go inside the `try` block, so that if it executes partially the `finally` will still restore the old loop.

asyncio/events.py
```diff
@@ -607,6 +607,27 @@ def new_event_loop(self):
 # Lock for protecting the on-the-fly creation of the event loop policy.
 _lock = threading.Lock()
 
+# A TLS for the current event loop, used by get_current_loop.
+_current_loop = threading.local()
```
gvanrossum: Why not use a subclass (like in BaseDefaultEventLoopPolicy)?

asyncio/test_utils.py
```diff
@@ -449,7 +449,13 @@ def new_test_loop(self, gen=None):
         self.set_event_loop(loop)
         return loop
 
+    def setUp(self):
+        self._get_current_loop = events._get_current_loop
+        events._get_current_loop = lambda: None
```
gvanrossum: Why not use mock.patch() as a class decorator? If I understand the docs correctly then you don't need to change everything to call super().setUp().

tests/test_events.py
```diff
@@ -2726,6 +2728,27 @@ def test_set_event_loop_policy(self):
...
 
+    def test_get_event_loop_returns_current_loop(self):
+        class Policy(asyncio.DefaultEventLoopPolicy):
+            def get_event_loop(self):
+                raise NotImplementedError
+
+        loop = None
+
+        old_policy = asyncio.get_event_loop_policy()
+        asyncio.set_event_loop_policy(Policy())
```
gvanrossum: Again, this ought to be inside the `try`.

asyncio/test_utils.py
1st1: `mock.patch` doesn't apply to subclasses of TestCase. So we'll need to decorate all TestCases, and we have a lot of them. IMO it's cleaner to fix `setUp` calls.

asyncio/base_events.py
1st1: Sure, will fix this.

asyncio/events.py
1st1: OK, will add a subclass. I guess avoiding a `getattr` call makes the code cleaner.

gvanrossum:
OK, nm on the mock.patch() thing then.

1st1:
@gvanrossum I've updated the PR.

gvanrossum:
What about my suggestion (in the previous PR, sorry) of using "active" or "running" instead of "current"? (I am coming back to this because in my head I keep missing the distinction between get_event_loop() and get_current_loop(). :-)

1st1:
I like `_get_running_loop`! "Running" is less ambiguous than "current" and "active".

gvanrossum:
Make it so!

1st1:
> Make it so!

Done :)  Please see the latest patch.

vxgmichel:
Looks good, better than the previous PR.

I have a question though. Why does `run_forever` store the old loop? It seems like this allows to run a loop inside another one. In the previous PR, setting a loop when another one is already running raises a `RuntimeError('A loop is already running')`. Would that be doable here?

asyncio/events.py
```diff
@@ -632,7 +656,17 @@ def set_event_loop_policy(policy):
 
 
 def get_event_loop():
-    """Equivalent to calling get_event_loop_policy().get_event_loop()."""
+    """Return an asyncio event loop.
+
+    When called from coroutines, this function will always return the
```
gvanrossum: I'd say "When called from a coroutine or a callback (e.g. scheduled with `call_soon` or a similar API),  this function ..."

gvanrossum:
@vxgmichel 
> I have a question though. Why does run_forever store the old loop? It seems like this allows to run a loop inside another one. In the previous PR, setting a loop when another one is already running raises a RuntimeError('A loop is already running'). Would that be doable here?

The more I think about it the more I agree. This would be one scenario where we have a chance to detect accidental blocking inside a coroutine or callback. @1st1 what do you think?

asvetlov:
Agree. Stackoverflow has several examples where authors tries to hide high level api under `loop.run_until_complete()`.
Explicit forbidding running another loop from coroutine is great idea.

gvanrossum:
(off-topic) We should have examples of a better recommended practice. IMO such high level APIs should fork a background thread to run the event loop and use concurrent.futures.Future instances to schedule calls and be able to wait for them.

vxgmichel:
@gvanrossum
> IMO such high level APIs should fork a background thread to run the event loop and use concurrent.futures.Future instances to schedule calls and be able to wait for them.

I have an [asyncio executor example](https://gist.github.com/vxgmichel/d16e66d1107a369877f6ef7e646ac2e5) that does pretty much that. I don't think it would fit in the official documentation, but maybe on [asyncio-doc](https://github.com/asyncio-doc/asyncio-doc)?

1st1:
> The more I think about it the more I agree. This would be one scenario where we have a chance to detect accidental blocking inside a coroutine or callback. @1st1 what do you think?

Yeah, let's do it. I'll update the PR.

1st1:
Updated.

gvanrossum:
PS. Next time can you not squash changes when you push to the PR/branch? It's easier to review what's new if local commits are preserved. We'll squash when merging the PR.

gvanrossum:
Oh nm. You didn't squash. GitHub lied. :-(

asyncio/base_events.py
```diff
@@ -394,13 +394,15 @@ def run_forever(self):
         self._check_closed()
         if self.is_running():
             raise RuntimeError('Event loop is running.')
```
gvanrossum: Would it be easy to change this error to "This event loop is already running." ?

gvanrossum:
LGTM! Go ahead and merge.

1st1:
Thank you for reviews @gvanrossum @vxgmichel @asvetlov!

gvanrossum:
PS. Why did the final squashed commit claim failing tests? I didn't see a Travis CI link.

1st1:
Short answer: race between Travis & GH (at least that's what it looks like).

Long answer: I squashed everything and force pushed it to the PR. When I merged it to the `master` branch. The PR was automatically closed by GH and it then aborted the builds (that were just started that time). Travis aborted the build marking it as a fail.

1st1:
The master looks green: https://travis-ci.org/python/asyncio/builds/173340363

gvanrossum:
Oh, your workflow is so different from mine... I use the web squash flow.

--Guido (mobile)

thehesiod:
this broke being able to use asyncio from forked processes created from a run loop :(  testcase:

```python
import asyncio
import multiprocessing
import zmq.asyncio


def sub_proc():
    async def doit():
        print("hello")

    loop = asyncio.new_event_loop()
    asyncio.set_event_loop(loop)
    loop.run_until_complete(doit())


async def main():
    proc = multiprocessing.Process(target=sub_proc)
    proc.start()
    proc.join()


if __name__ == '__main__':
    _loop = asyncio.get_event_loop()
    _loop.run_until_complete(main())
```

this works in python3.5.2 but fails in 3.5.3

1st1:
Forking from a running event loop triggers many undefined behaviours. Just don't do it.

thehesiod:
this is a really unfortunate response as this is a common workflow for async unittests (https://github.com/Martiusweb/asynctest) and would require ALL workers to be created before your first async method, which for things like pools cannot be guaranteed.  Further, it would break scenarios where you have a sub-processes re-spawned after serving x number of requests  I've been using this workflow for over a year in a production environment without issue.  Unless the point of this PR was to break this workflow I think it should addressed.  What this PR is saying is that async processes cannot launch async sub-processes, which was never stated before.

Further, in general, async forking support has been mothballed for more than a year and really needs to be addressed per the issue I referenced above.

1st1:
> this is a really unfortunate response as this is a common workflow for async unittests (https://github.com/Martiusweb/asynctest) and would require ALL workers to be created before your first async method, which for things like pools cannot be guaranteed. 

If one package does something in some particular way, it doesn't mean that it's a correct way. It means that we need to fix that one package.

> [..]  What this PR is saying is that async processes cannot launch async sub-processes, which was never previously stated or enforced.

No, it doesn't say that. It just happened to expose a bug (forking from a running event loop was never officially supported). Forking with a running event loop is discouraged in virtually all async frameworks. There are dozens of side effects that can't be reliably resolved in a cross-platform case. Alternative asyncio event loop implementations such as uvloop will simply segfault if you do this.

There is a way to fork reliably, it just requires some boilerplate code. The key is to fork with the loop being **paused**. For example (untested):

```python
fork_cb = None

def schedule_fork(new_proc):
    global fork_cb
    fork_cb = new_proc
    asyncio.get_event_loop().stop()

async def coro():
    # some asyncio code...
    # some asyncio code...
    # some asyncio code...

    def new_proc():
        print('forked process', os.getpid())

    print('forking from', os.getpid())
    schedule_fork(new_proc)


while True:
    asyncio.run_forever()
    if fork_cb is not None:
        cb, fork_cb = fork_cb, None
        if not os.fork():
            cb()
    else:
        break
```

A workaround would be to store PID in the threadlocal object that `get_event_loop()` is using. But that would make the call slower and will only mask the real issue and enable bad practices.

> Further, in general, async forking support has been mothballed for more than a year and really needs to be addressed per the issue I referenced above.

I think we actually can add built-in support for forking to asyncio: a coroutine `asyncio.fork(cb)`:

```python
async def some_coro():
    await asyncio.fork(code_for_child_process)
```

I'll experiment with the idea and maybe propose this in 3.7.

1st1:
@Martiusweb Would you be able to fix forking in asynctest by using the approach outlined in the above  ^^ comment?

thehesiod:
FYI another example that just uses asyncio to fork:
```python
import asyncio
from concurrent.futures import ProcessPoolExecutor


def sub_proc():
    async def doit():
        print("hello")

    loop = asyncio.new_event_loop()
    asyncio.set_event_loop(loop)
    loop.run_until_complete(doit())


async def main():
    ppool = ProcessPoolExecutor()
    await loop.run_in_executor(ppool, sub_proc)


if __name__ == '__main__':
    _loop = asyncio.get_event_loop()
    _loop.run_until_complete(main())
```

I think the justification needs to be made if python's base modules are fork safe or not.  If asyncio is not fork safe this seems to go against the other python base modules.  What work is required to be able to subprocess the BaseEventLoop to make it fork safe?

1st1:
> IMHO I believe the workaround you mentioned is not robust in that one may not control the main run loop (like when running a aiohttp server) .. 

> .. In my example I run an aiohttp server that dynamically launches processes from a worker-pool to process tens of thousands of S3 requests.

Forking from within a running HTTP server coroutine is a very bad idea. It's a blocking operation that might take non-trivial amount of time. You better pre-fork, or have a master process that controls its children processes and creates new ones when needed.

You can also use `loop.subprocess_exec` to safely run your S3 logic.

> and further I worry of the heavy hand of stopping the event loop for complicated applications which have many tasks which are blocked waiting for input, timers, and tasks waiting of various events.

Event loop can safely be stopped and resumed, no matter how many timers/tasks you have. If you care about timeouts being triggered because you pause the loop to fork -- the same will happen when you use `os.fork()` (again, it's not a cheap-n-fast operation).

> IMO if uvloop crashes that would be a problem in its run-loop impl and should be addressed separately and not affect the BaseEvent loop impl.

True (and that segfault will soon be fixed).

**But** there are problems even with forking pure python asyncio programs. For instance, `epoll` is fundamentally not fork-safe. If you continue to use the same event loop in the forked child process you will encounter bugs/crashes/wrong behaviour.

One solution is to use the `multiprocessing` module, which is supported by both asyncio and uvloop, or to do fork+exec manually.

> FYI another example that just uses asyncio to fork:
> [..] ProcessPoolExecutor example [..]

You can also use `run_in_executor` API with process pools -- that is also fully supported (because `concurrent.futures` uses `multiprocessing`).

Calling bare `os.fork()` is fundamentally unsafe, you simply should not use it. It's a low level syscall, and asyncio event loop is at least one level above it. What you are asking is to fix one particular aspect of the API, so that you can continue to use what worked by accident. But even if we make `get_event_loop` work you will eventually have other problems with `os.fork`.

gevent, for instance, monkey patches `os.fork` to make it work safe. As I said in my previous comment, we might want to add a specialized `asyncio.fork` method, but bare `os.fork` is very unlikely to to be fully supported ever.

> I think the justification needs to be made if python's base modules are fork safe or not. If asyncio is not fork safe this seems to go against the other python base modules. What work is required to be able to subprocess the BaseEventLoop to make it fork safe?

Strictly speaking any network application (blocking or non-blocking) is not `os.fork` friendly. You have to do the forking with extra care, and generally people only do fork+exec. The officially recommended way to do multiprocessing is to use the `multiprocessing` package.

> You also say forking from async from run-loop was never officially supported, but by the same argument it was never officially not-supported either from what I gather from the docs (https://docs.python.org/3/library/asyncio-eventloop.html), and it worked.

True, this is something we will hopefully fix very soon.

1st1:
> thanks for the insight, but I think our streams aren't quite crossing yet :) I never stated the workflows involved directly forking, I gave two examples that this PR broke without using fork directly. [..]

Right. I tried your script from https://github.com/python/asyncio/pull/452#issuecomment-281811666 and indeed what should work doesn't work! Seems like it's a bug after all, I'll open a PR.

thehesiod:
yay =)  ya it's the same problem as the first example.

PR #453, opened 2016-11-06
Make loop TCP APIs accept only TCP streaming sockets

1st1:
Currently, `loop.create_server`, `loop.create_connection`, and `loop.connect_accepted_socket` accept any kind of socket, although they are documented (and used by asyncio users) as TCP API.

Allowing those methods to accept any kinds of sockets leads to hard to debug bugs.  For instance, aiohttp gunicorn worker simply passes all sockets to `create_server` (including UNIX sockets). We even have a unittest in asyncio that does that: `test_create_datagram_endpoint_sock` passes a UDP socket to `create_connection`.

TCP, UDP and UNIX sockets are different from each other and have some subtle API differences. We shouldn't allow to blindly pass any of them to `create_server` and other TCP APIs.

### DIFF
--- a/asyncio/base_events.py
+++ b/asyncio/base_events.py
@@ -84,12 +84,26 @@ def _set_reuseport(sock):
                              'SO_REUSEPORT defined but not implemented.')
 
 
-# Linux's sock.type is a bitmask that can include extra info about socket.
-_SOCKET_TYPE_MASK = 0
-if hasattr(socket, 'SOCK_NONBLOCK'):
-    _SOCKET_TYPE_MASK |= socket.SOCK_NONBLOCK
-if hasattr(socket, 'SOCK_CLOEXEC'):
-    _SOCKET_TYPE_MASK |= socket.SOCK_CLOEXEC
+def _is_stream_socket(sock):
+    # Linux's socket.type is a bitmask that can include extra info
+    # about socket, therefore we can't do simple
+    # `sock_type == socket.SOCK_STREAM`.
+    return (sock.type & socket.SOCK_STREAM) == socket.SOCK_STREAM
+
+
+def _is_dgram_socket(sock):
+    # Linux's socket.type is a bitmask that can include extra info
+    # about socket, therefore we can't do simple
+    # `sock_type == socket.SOCK_DGRAM`.
+    return (sock.type & socket.SOCK_DGRAM) == socket.SOCK_DGRAM
+
+
+def _is_ip_socket(sock):
+    if sock.family == socket.AF_INET:
+        return True
+    if hasattr(socket, 'AF_INET6') and sock.family == socket.AF_INET6:
+        return True
+    return False
 
 
 def _ipaddr_info(host, port, family, type, proto):
@@ -102,8 +116,12 @@ def _ipaddr_info(host, port, family, type, proto):
             host is None:
         return None
 
-    type &= ~_SOCKET_TYPE_MASK
     if type == socket.SOCK_STREAM:
+        # Linux only:
+        #    getaddrinfo() can raise when socket.type is a bit mask.
+        #    So if socket.type is a bit mask of SOCK_STREAM, and say
+        #    SOCK_NONBLOCK, we simply return None, which will trigger
+        #    a call to getaddrinfo() letting it process this request.
         proto = socket.IPPROTO_TCP
     elif type == socket.SOCK_DGRAM:
         proto = socket.IPPROTO_UDP
@@ -124,7 +142,9 @@ def _ipaddr_info(host, port, family, type, proto):
             return None
 
     if family == socket.AF_UNSPEC:
-        afs = [socket.AF_INET, socket.AF_INET6]
+        afs = [socket.AF_INET]
+        if hasattr(socket, 'AF_INET6'):
+            afs.append(socket.AF_INET6)
     else:
         afs = [family]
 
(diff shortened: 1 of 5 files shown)


tests/test_events.py
```diff
@@ -1377,6 +1377,11 @@ def datagram_received(self, data, addr):
         server.transport.close()
 
     def test_create_datagram_endpoint_sock(self):
+        if (sys.platform == 'win32' and
+                isinstance(self.loop, proactor_events.BaseProactorEventLoop)):
+            raise unittest.SkipTest(
+                'UCP not supported with proactor event loops')
```
gvanrossum: Typo, UDP?

tests/test_events.py
```diff
@@ -1394,7 +1399,7 @@ def test_create_datagram_endpoint_sock(self):
         else:
             assert False, 'Can not create socket.'
 
-        f = self.loop.create_connection(
+        f = self.loop.create_datagram_endpoint(
```
gvanrossum: I'm worried here that this test doesn't test much...

1st1: Yeah, you're right, it tests only that the method doesn't crash. Still it was testing a wrong method.

asyncio/base_events.py
```diff
@@ -1027,6 +1031,9 @@ def create_server(self, protocol_factory, host=None, port=None,
         else:
             if sock is None:
                 raise ValueError('Neither host/port nor sock were specified')
+            if sock.family not in {socket.AF_INET, socket.AF_INET6}:
```
gvanrossum: Careful, check line 972/976. AF_INET6 may not always exist in the socket module (apparently).

asyncio/base_events.py
```diff
@@ -1048,6 +1055,13 @@ def connect_accepted_socket(self, protocol_factory, sock, *, ssl=None):
         This method is a coroutine.  When completed, the coroutine
         returns a (transport, protocol) pair.
         """
+        socket_families = {socket.AF_INET, socket.AF_INET6}
+        if hasattr(socket, 'AF_UNIX'):
+            socket_families.add(socket.AF_UNIX)
```
gvanrossum: If a UNIX socket is fine here, why not in create_connection()? They both end up calling `_create_connection_transport()`.

asyncio/base_events.py
1st1: Good point. Will fix this.

asyncio/base_events.py
1st1: Yes, in asyncio both `create_connection` and `connect_accepted_socket` use similar internal API. It's not the same in uvloop.

The problem is that we have `create_connection` and `create_unix_connection`, that are designed for TCP and UNIX sockets.  uvloop implements this spec, using appropriate low-level libuv primitives for each case.  It doesn't make a lot of sense to add "if" checks to accept Unix sockets in `create_connection` (as we already have *dedicated* APIs for that).

`connect_accepted_socket` on the other hand doesn't specify the kind of socket it accepts. And it doesn't make a lot of sense to add `connect_accepted_unix_socket`.

1st1: That means that users will be able to pass `AF_BLUETOOTH` sockets or something.  Those aren't supported in uvloop, and I'm not sure that even asyncio supports them properly.

In any case, I think you're right, let's only check for `SOCK_STREAM` here. I'll also add `SOCK_STREAM` checks in `create_connection`, `create_server`, as they simply won't work with anything else. `AF_INET6` has to be properly handled too.  Let me update the PR.

1st1:
> (OT: connect_accepted_socket seems undocumented.)

I've committed Jim's documentation patch: http://bugs.python.org/issue27392. I've also updated this PR to address your comments.

asyncio/base_events.py
gvanrossum: I have some additional worry now. You've mentioned twice (I think) "but uvloop doesn't support that". I don't think that's a valid argument. (Haven't we discussed this before?) I think it's fine if uvloop doesn't support e.g. bluetooth sockets but pure-Python asyncio does. That should be documented with uvloop; but I don't think "uvloop can't do this" is enough to conclude "hence asyncio shouldn't do this either". (Especially since I remember seeing a bug about code that would crash when passing a bluetooth address; IIRC it was a getaddrinfo() issue.)

1st1: Re uvloop - I agree. The updated PR only makes sure that:

1. `create_datagram_endpoint` only accepts `SOCK_DGRAM` sockets (it doesn't make sense to accept anything else).

2. `create_connection` and `create_server` only accept `AF_INET`, `AF_INET6` and `SOCK_STREAM` sockets.  We have separate asyncio API for `AF_UNIX` sockets.

3. `connect_accepted_socket` only accepts `SOCK_STREAM` sockets (no restrictions on family, event `AF_BLUETOOTH` will be accepted just fine. Even though uvloop won't work with the latter).

All in all, this PR is not about uvloop, it only ensures that asyncio APIs are used correctly. It doesn't limit asyncio in any way.

gvanrossum:
Thanks, LGTM now!

1st1:
I've updated the patch with more unittests and more robust checks of `socket.type`. Will merge it later.

asyncio/base_events.py
```diff
@@ -84,12 +84,26 @@ def _set_reuseport(sock):
                              'SO_REUSEPORT defined but not implemented.')
 
 
-# Linux's sock.type is a bitmask that can include extra info about socket.
-_SOCKET_TYPE_MASK = 0
-if hasattr(socket, 'SOCK_NONBLOCK'):
-    _SOCKET_TYPE_MASK |= socket.SOCK_NONBLOCK
-if hasattr(socket, 'SOCK_CLOEXEC'):
-    _SOCKET_TYPE_MASK |= socket.SOCK_CLOEXEC
+def _is_stream_socket(sock_type):
```
gvanrossum: I would rename this to `_is_stream_socket_type()`, and similar for the following functions (using `_family` as the suffix for `_is_ip_socket`) to emphasize which field of the socket is supposed to be passed into these functions.

gvanrossum: Even better!

1st1:
Thanks for the review, Guido. I've implemented the same set of features in uvloop, and will watch closely if they cause any problems (I highly doubt that).

gvanrossum:
Whee!

1st1:
I rolled back the restriction of passing `AF_UNIX` to `create_connection` and `create_server`. Turns out too many people rely on that, so I guess we can just support that (it's easy to fix uvloop to support this as well).

Issue #396, opened 2016-08-05
TypeError: signal handler must be signal.SIG_IGN, signal.SIG_DFL, or a callable object


sametmax:
If you register a signal handler for SIGTERM, start the event loop then kill it with CTRL + C, you get this error:

```
Exception ignored in: <bound method BaseEventLoop.__del__ of <_UnixSelectorEventLoop running=False closed=True debug=False>>
Traceback (most recent call last):
  File "/usr/lib/python3.5/asyncio/base_events.py", line 431, in __del__
  File "/usr/lib/python3.5/asyncio/unix_events.py", line 58, in close
  File "/usr/lib/python3.5/asyncio/unix_events.py", line 140, in remove_signal_handler
  File "/usr/lib/python3.5/signal.py", line 47, in signal
TypeError: signal handler must be signal.SIG_IGN, signal.SIG_DFL, or a callable object
```

Minimal code to get the error:

```
import signal
import asyncio

loop = asyncio.get_event_loop()

def bar():
    loop.stop()

loop.add_signal_handler(signal.SIGTERM, bar)
loop.run_forever()
```

This will not occur if you register a handler for another signal, such as SIGINT.

vxgmichel:
> This will not occur if you register a handler for another signal, such as SIGINT.

It does actually occur for all signals except `SIGINT` (`SIGINT` has a special treatment in `loop.remove_signal_handler`). This error has been reported on bugs.python.org as [issue32548](http://bugs.python.org/issue23548). I've looked into it, and it turns out this problem is unrelated to asyncio. It can be reproduced using the `_signal` module only:

```python
import _signal

class A:
    def __del__(self):
        _signal.signal(_signal.SIGTERM, _signal.SIG_DFL)

a = A()
```

It also affects python 2. This happens because the `signalmodule` resources are [cleared](https://hg.python.org/cpython/file/default/Modules/signalmodule.c#l1532) before the garbage collector calls `a.__del__`. It causes [this `if` statement](https://hg.python.org/cpython/file/default/Modules/signalmodule.c#l435) to fail and a misleading `TypeError` is raised.

It's quite similar to [issue 14173](http://bugs.python.org/issue14173) (fixed in 2012).

sametmax:
Thanks. Let me close this then, no need for duplicate tickets.

PR #456, opened 2016-11-07
Fix remove_signal_handler to not to crush after PyOS_FiniInterrupts

1st1:
I think I've figured out what's going on with `remove_signal_handler`.

This PR fixes:

* https://github.com/python/asyncio/issues/396
* http://bugs.python.org/issue23548
* http://bugs.python.org/issue28628

### DIFF
--- a/asyncio/unix_events.py
+++ b/asyncio/unix_events.py
@@ -1,5 +1,6 @@
 """Selector event loop for Unix with signal handling."""
 
+import atexit
 import errno
 import os
 import signal
@@ -9,6 +10,7 @@ import subprocess
 import sys
 import threading
 import warnings
+import weakref
 
 
 from . import base_events
@@ -39,6 +41,12 @@ def _sighandler_noop(signum, frame):
     pass
 
 
+def _loop_atexit_callback(loop_ref):
+    loop = loop_ref()
+    if loop is not None:
+        loop._interpreter_shutting_down = True
+
+
 class _UnixSelectorEventLoop(selector_events.BaseSelectorEventLoop):
     """Unix event loop.
 
@@ -47,6 +55,15 @@ class _UnixSelectorEventLoop(selector_events.BaseSelectorEventLoop):
 
     def __init__(self, selector=None):
         super().__init__(selector)
+
+        # atexit callbacks are fired before PyOS_FiniInterrupts, which
+        # allows us to workaround bugs in remove_signal_handler.
+        atexit.register(_loop_atexit_callback, weakref.ref(self))
+        # When _interpreter_shutting_down is True, PyOS_FiniInterrupts
+        # has already been called and signalmodule's internal state was
+        # cleaned up.
+        self._interpreter_shutting_down = False
+
         self._signal_handlers = {}
 
     def _socketpair(self):
@@ -124,6 +141,11 @@ class _UnixSelectorEventLoop(selector_events.BaseSelectorEventLoop):
 
         Return True if a signal handler was removed, False if not.
         """
+        if self._interpreter_shutting_down:
+            # The interpreter is being shutdown.  `PyOS_FiniInterrupts`
+            # was already called and it has restored all signals already.
+            return
+
         self._check_signal(sig)
         try:
             del self._signal_handlers[sig]


asyncio/unix_events.py
```diff
@@ -124,6 +141,11 @@ def remove_signal_handler(self, sig):
 
         Return True if a signal handler was removed, False if not.
         """
+        if self._interpreter_shutting_down:
+            # The interpreter is being shutdown.  `PyOS_FiniInterrupts`
+            # was already called and it has restored all signals already.
+            return
```
gvanrossum: This should return a bool.

gvanrossum: Only in debug mode. But frankly I don't understand this code.

gvanrossum:
Do we know that the atexit handler will always be called early enough?

1st1:
> I don't understand all this well enough to know whether this is a good idea or not. We really need to find someone else who wants to co-own asyncio -- ATM I feel my role is mostly that of giving you permission to commit what you see fit and I am not actually comfortable with this -- we'll end up with there being only one person who understands asyncio and that's not enough for such an important stdlib module.

I agree. I'm myself not feeling comfortable, even two of us doesn't seem to be enough to support asyncio.

@asvetlov, @methane  and @haypo: are you interested to help us with asyncio? At least to offer help with PRs and participate in related discussions.

> Do we know that the atexit handler will always be called early enough?

`Py_FinalizeEx` first calls atexit functions and only after that it calls `PyOS_FiniInterrupts()`. The latter cleans up the internal state of `signals` module, making it useless. So yes, we're confident about atexit.

asvetlov:
I think the PR makes sense: we raises `ResourceWarning` for unclosed transports in debug mode.
I believe we should do the same for unclosed loop too.

gvanrossum:
OK, then LGTM.

vxgmichel:
Shouldn't it be fixed in `signalmodule.c` instead? (see my [comment](https://github.com/python/asyncio/issues/396#issuecomment-237863433) in PR #396)

gvanrossum:
If something needs to be changed in signalmodule.c please open an issue on
bugs.python.org.

1st1:
Yes, the bug in _signal can (and should) be fixed, and I have a working patch for that. My worry is that it might be a bit too late to do that for 3.6. I'll submit the patch and ask Ned's opinion.

1st1:
I've no idea what's going on. I can reproduce this bug with python installed from MacPorts. I can't reproduce it with python I build from source.

1st1:
The bug is the following program:

```python
import signal
import asyncio

def foo():
    pass

async def bar():
    pass

def main():
    loop.add_signal_handler(signal.SIGINT, lambda: None)
    loop.add_signal_handler(signal.SIGHUP, lambda: None)
    loop.run_until_complete(bar())

if __name__ == "__main__":
    main()
```

It should spit out something like this: 
```
Exception ignored in: <bound method BaseEventLoop.__del__ of <_UnixSelectorEventLoop running=False closed=True debug=False>>
Traceback (most recent call last):
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/asyncio/base_events.py", line 431, in __del__
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/asyncio/unix_events.py", line 58, in close
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/asyncio/unix_events.py", line 139, in remove_signal_handler
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/signal.py", line 47, in signal
TypeError: signal handler must be signal.SIG_IGN, signal.SIG_DFL, or a callable object
```

gvanrossum:
Maybe they ran configure with different parameters or in a different
context or with a different version of Xcode?

1st1:
Maybe. It should be something that changes how & when objects are GCed. The above traceback happens when objects are GCed after the signals module is finalized (and that's what I've seen in reports from asyncio and uvloop users). I'll try to get to the bottom of this.

vxgmichel:
@1st1
I tried to reproduce the bug with python 3.5 and python 3.6 (built from the latest sources) with your test for `asyncio` and this one for `signal`:

```
import _signal

class A:
    def __del__(self):
        _signal.signal(_signal.SIGTERM, _signal.SIG_DFL)

a = A()
```
My results:

           | asyncio test | signal test
-----------|--------------|-------------
python 3.5 | TypeError    | TypeError
python 3.6 | No Error     | TypeError

PR #457, opened 2016-11-09
Make __format_callbacks slightly faster

illia-v:


1st1:
Thanks for trying to improve the code!

In 3.6 we re-implemented Futures in C, and this code has been changed. If you want to see if your optimization still applies, please check that out and submit a patch on bugs.python.org. Closing this PR.

Issue #404, opened 2016-08-26
Don't pass ssl certivicate error to loop.call_exception_handler()


asvetlov:
The following code prints uncaught handler exception but shouldn't
```
import asyncio
import contextlib
import aiohttp

async def main(url):
    async with aiohttp.ClientSession() as session:
        with contextlib.suppress(Exception):
            async with session.get(url):
                await response.read()

url = 'https://wrong.host.badssl.com'
loop = asyncio.get_event_loop()
loop.run_until_complete(main(url))
```
Exception is raised, everything is correct except printing non-catched exception in SSL protocol callback.

With `PYTHONASYNCIODEBUG` enabled the output is:
```
<asyncio.sslproto.SSLProtocol object at 0x7f10e2212470>: SSL handshake failed on verifying the certificate
Traceback (most recent call last):
  File "/usr/lib/python3.5/asyncio/sslproto.py", line 568, in _on_handshake_complete
    raise handshake_exc
  File "/usr/lib/python3.5/asyncio/sslproto.py", line 201, in feed_ssldata
    self._sslobj.do_handshake()
  File "/usr/lib/python3.5/ssl.py", line 638, in do_handshake
    match_hostname(self.getpeercert(), self.server_hostname)
  File "/usr/lib/python3.5/ssl.py", line 297, in match_hostname
    % (hostname, ', '.join(map(repr, dnsnames))))
ssl.CertificateError: hostname 'wrong.host.badssl.com' doesn't match either of '*.badssl.com', 'badssl.com'
Exception in callback None
handle: <Handle cancelled _SelectorSocketTransport._read_ready() created at /usr/lib/python3.5/asyncio/selector_events.py:232>
source_traceback: Object created at (most recent call last):
  File "3.py", line 13, in <module>
    loop.run_until_complete(main(url))
  File "/usr/lib/python3.5/asyncio/base_events.py", line 375, in run_until_complete
    self.run_forever()
  File "/usr/lib/python3.5/asyncio/base_events.py", line 345, in run_forever
    self._run_once()
  File "/usr/lib/python3.5/asyncio/base_events.py", line 1304, in _run_once
    handle._run()
  File "/usr/lib/python3.5/asyncio/events.py", line 125, in _run
    self._callback(*self._args)
  File "/usr/lib/python3.5/asyncio/selector_events.py", line 232, in add_reader
    handle = events.Handle(callback, args, self)
Traceback (most recent call last):
[...]

PR #460, opened 2016-11-11
Fix for #404

asvetlov:
Don't pass ssl certivicate error to loop.call_exception_handler()
Handle ssl.CertificateError only because ssl.SSLError may be raised not only on connection establishment.

@1st1 please take a look

### DIFF
--- a/asyncio/base_events.py
+++ b/asyncio/base_events.py
@@ -37,6 +37,11 @@ from . import tasks
 from .coroutines import coroutine
 from .log import logger
 
+try:
+    import ssl
+except ImportError:
+    ssl = None
+
 
 __all__ = ['BaseEventLoop']
 
@@ -53,6 +58,8 @@ _MIN_CANCELLED_TIMER_HANDLES_FRACTION = 0.5
 # methods (_fatal_error())
 _FATAL_ERROR_IGNORE = (BrokenPipeError,
                        ConnectionResetError, ConnectionAbortedError)
+if ssl is not None:
+    _FATAL_ERROR_IGNORE = _FATAL_ERROR_IGNORE + (ssl.CertificateError,)
 
 
 def _format_handle(handle):


asvetlov:
Fixes #404

1st1 (changes requested):
Updating `_FATAL_ERROR_IGNORE` worries me: I don't think I understand all consequences of this PR.  Could you please elaborate more?

asvetlov:
`_FATAL_ERROR_IGNORE` is a tuple of exceptions which doesn't generate noisy `loop.call_exception_handler()` if connection cannot be established (see https://github.com/python/asyncio/blob/master/asyncio/selector_events.py#L633) for usage example.
`ssl.CertificateError` goes to this category -- it could be raised only at connecting stage.
The exception still is thrown to outer code but not logged unless debug mode is disabled.

1st1:
Makes sense. Would you be able to write a unittest for this?

1st1:
@asvetlov If you want this in 3.6.1 you need to add a unittest and resubmit this PR to python/cpython today.

AraHaan:
@1st1 ok, so 3.6.1 is going out after today? btw would it also include the latest version of zlib as well because it has updated a to version above 1.2.8 this year to fix bugs and stuff.

1st1:
> @1st1 ok, so 3.6.1 is going out after today? btw would it also include the latest version of zlib as well because it has updated a to version above 1.2.8 this year to fix bugs and stuff.

Yes. Re zlib - not sure, please check on GH or bugs.python.org or python-dev mailing list.

AraHaan:
hmm, from the looks on http://svn.python.org/projects/external/ I only see it having zlib v1.2.3 But I could have sworn python 3.6.0 was using 1.2.8. Will look at bugs.python.org or the python-dev irc channel.

AraHaan:
@1st1 It looks like this repo was not marked as ``Read Only`` in it's settings to disable (I think it disables) the new issue / Pull Request buttons. Might want to do it explicitly in case someone does not read the readme in the redirect branch.

decaz:
@asvetlov are you going to resubmit this PR to python/cpython ? I hope it's not abandoned.

PR #462, opened 2016-11-12
Fix create_unix_server to support Path-like objects (PEP 519)

1st1:
One of the recent commits broke support of PEP 519. This PR fixes `loop.create_unix_server` to support Path-like objects + a regression test.

cc @brettcannon

### DIFF
--- a/asyncio/unix_events.py
+++ b/asyncio/unix_events.py
@@ -39,6 +39,13 @@ def _sighandler_noop(signum, frame):
     pass
 
 
+try:
+    _fspath = os.fspath
+except AttributeError:
+    # Python 3.5 or earlier
+    _fspath = lambda path: path
+
+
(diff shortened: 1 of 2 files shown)


asyncio/unix_events.py
```diff
@@ -39,6 +39,13 @@ def _sighandler_noop(signum, frame):
     pass
 
 
+try:
+    _fspath = os.fspath
+except AttributeError:
+    # Python 3.5 or earlier
+    _fspath = lambda path: path
```
asvetlov: Maybe support `pathlib.Path` for Python 3.5 also?
It could be implemented by `isinstance()` check.

asvetlov: ok

1st1:
Pushed in https://github.com/python/asyncio/commit/fff05d480760703adbc3e2d4cb3dbcfbff803c29

Issue #363, opened 2016-06-22
NameError in  _fatal_error


rwalkerands:
In `asyncio/sslproto.py`, method `_fatal_error`, the use of `base_events._FATAL_ERROR_IGNORE` (currently at https://github.com/python/asyncio/blob/master/asyncio/sslproto.py#L658) causes a `NameError`.

Need to add this to the top:

    from . import base_events

In my case, I'm seeing this error because of an underlying problem in my code (which I've now fixed -- I had erroneously left open a socket, and this happens when the socket is later cleaned up), but this error made it just a little harder to diagnose the cause.

See https://groups.google.com/forum/?fromgroups#!searchin/python-tulip/nameerror/python-tulip/RbZ_qWuqSZI/NfsqNvA_CwAJ and https://groups.google.com/forum/#!searchin/linux.debian.bugs.dist/827453/linux.debian.bugs.dist/6aLdMUvvSqw/YSKbGZzyEQAJ for examples of someone else getting the same symptom.

gvanrossum:
Yup. Care to send us a PR with a fix? It should have a test.

rwalkerands:
The first example from the forum post by Jack Bates is enough to generate the NameError:

    import asyncio
    asyncio.get_event_loop().run_until_complete(
        asyncio.open_connection('google.com', 'https', ssl=True))

My difficulty in writing a test based on this is how to make assertions about the presence/absence of a particular exception that seems to be generated during the cleanup (gc?) at program _exit_. I've been unable to force the cleanup in such a way as to trigger the exception.

rwalkerands:
I guess one could just call `_fatal_error()` directly:

    import asyncio
    import unittest

    class TestTransport(unittest.TestCase):

        def test_fatal_error_NameError(self):
            reader, writer = asyncio.get_event_loop().run_until_complete(
                asyncio.open_connection('google.com', 'https', ssl=True))
            writer.transport._ssl_protocol._fatal_error(None)

    if __name__ == '__main__':
        unittest.main()

Output before fixing:

    E
    ======================================================================
    ERROR: test_fatal_error_NameError (__main__.TestTransport)
    ----------------------------------------------------------------------
    Traceback (most recent call last):
      File "script4.py", line 9, in test_fatal_error_NameError
        writer.transport._ssl_protocol._fatal_error(None)
      File "/home/users/u9002369/py352rc1/lib/python3.5/asyncio/sslproto.py", line 658, in _fatal_error
[...]

PR #467, opened 2016-11-16
Cancelled lock waiter wakes up the next one if any

msornay:
Fixes http://bugs.python.org/issue27585

Without the fix the unit test deadlock, I haven't managed to make it fail (using wait_for() seems to avoid the deadlock)

### DIFF
--- a/tests/test_pep492.py
+++ b/tests/test_pep492.py
@@ -71,6 +71,31 @@ class LockTests(BaseTest):
             self.loop.run_until_complete(test(primitive))
             self.assertFalse(primitive.locked())
 
+    def test_finished_waiter_cancelled(self):
+
+        async def create_waiter(lock, fut):
+            fut.set_result(True)
+            async with lock:
+                pass
+
+        async def runner():
+            lock = asyncio.Lock(loop=self.loop)
+
+            await lock.acquire()
+
+            fut = self.loop.create_future()
+            task = asyncio.ensure_future(create_waiter(lock, fut),
+                                             loop=self.loop)
+            await fut
+
+            lock.release()
+            task.cancel()
+
+            async with lock:
+                pass
+
+        self.loop.run_until_complete(runner())
+
 
 class StreamReaderTests(BaseTest):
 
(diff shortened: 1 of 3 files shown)


1st1:
@gvanrossum @asvetlov I'm not sure I can review this (I never touched/used that code). Can you take a look if you have time and you are familiar with the code?

gvanrossum:
Sorry, I can't do it this week or next.

tests/test_pep492.py
```diff
@@ -71,6 +71,31 @@ def test_context_manager_with_await(self):
             self.loop.run_until_complete(test(primitive))
             self.assertFalse(primitive.locked())
 
+    def test_finished_waiter_cancelled(self):
+
+        async def create_waiter(lock, fut):
```
1st1: Please use the `@coroutine` decorator and generators in unittests. We'll start using 'async def' when 3.6 is out and we start working on Python 3.7.

1st1:
Looks like this PR is doing a right thing. @msornay please update the tests, and I will merge it.

tests/test_pep492.py
msornay: Even in `test_pep492.py` ? Should I move the test ?

the-knights-who-say-ni:
Hello, and thanks for your contribution!

Unfortunately we couldn't find an account corresponding to your GitHub username at [bugs.python.org](http://bugs.python.org/) (b.p.o). If you don't already have an account at b.p.o, please [create one](http://bugs.python.org/user?@template=register) and make sure to add your GitHub username. If you do already have an account at b.p.o then please go there and under "Your Details" add your GitHub username.

And in case you haven't already, please make sure to sign the [PSF contributor agreement](https://www.python.org/psf/contrib/contrib-form/) (CLA); we can't legally look at your contribution until you have signed the CLA.

Once you have done everything that's needed, please reply here and someone will verify everything is in order.

msornay:
Everything should be in order.

pklanke:
After applying patch #363 and #467 to Python 3.5.2, the dead-lock did indeed no longer occur, but the opposite did; Sometimes on releasing the lock, a RuntimeError exception is raised telling me the lock was not acquired.

lock.acquire()
try:
    ... do stuff ...
finally:
    lock.release() # RuntimeError occured here

So far, I did not manage to create a simple test case, and I know this does not help much without it, but should someone experience this same behavior, then just know you are not the only one.

1st1:

PR #485, opened 2017-01-13
Delay closing SubprocessStreamProtocol's Transport until all pipes are closed.

### LINKED ISSUE #484: [BUG] subprocess._loop deleted too soon causing exception when trying to read from pipe
From https://bugs.python.org/issue23242#msg284930 

The following script is used to reproduce the bug:
```python
import asyncio

async def execute():
    process = await asyncio.create_subprocess_exec(
        "timeout", "0.1", "cat", "/dev/urandom", stdout=asyncio.subprocess.PIPE)

    while True:
        data = await process.stdout.read(65536)
        print('read %d bytes' % len(data))
        if data:
            await asyncio.sleep(0.3)
        else:
            break

asyncio.get_event_loop().run_until_complete(execute())
```
will produce following output and terminate with exception:
```
read 65536 bytes
read 65536 bytes
Traceback (most recent call last):
  File "read_subprocess.py", line 18, in <module>
    asyncio.get_event_loop().run_until_complete(execute())
  File "/usr/lib/python3.6/asyncio/base_events.py", line 466, in run_until_complete
    return future.result()
  File "read_subprocess.py", line 9, in execute
    data = await process.stdout.read(65536)
  File "/usr/lib/python3.6/asyncio/streams.py", line 634, in read
    self._maybe_resume_transport()
  File "/usr/lib/python3.6/asyncio/streams.py", line 402, in _maybe_resume_transport
    self._transport.resume_reading()
  File "/usr/lib/python3.6/asyncio/unix_events.py", line 401, in resume_reading
    self._loop._add_reader(self._fileno, self._read_ready)
AttributeError: 'NoneType' object has no attribute '_add_reader'
```

When the process exits https://github.com/python/asyncio/blob/master/asyncio/unix_events.py#L444 is called which sets this._loop = None
Next time read() is called on the pipe the above exception is thrown.
I have tried to fix this issue myself but would sometimes have read terminate too early and miss the last chunks of data.

sethmlarson:
I believe this change fixes #484.

Delays the closing for `SubprocessStreamProtocol._transport` until all pipes have been closed and the process has exited rather than just the process exiting. This allows for reading data from the subprocess that is still pending in the pipe after the process has exited.

I haven't run tests or created a test for this exact issue but I will create a few tests later.

### DIFF
--- a/asyncio/subprocess.py
+++ b/asyncio/subprocess.py
@@ -24,7 +24,8 @@ class SubprocessStreamProtocol(streams.FlowControlMixin,
         self._limit = limit
         self.stdin = self.stdout = self.stderr = None
         self._transport = None
-        self._pipe_fds = [-1]
+        self._process_exited = False
+        self._pipe_fds = []
 
     def __repr__(self):
         info = [self.__class__.__name__]
@@ -88,15 +89,17 @@ class SubprocessStreamProtocol(streams.FlowControlMixin,
                 reader.feed_eof()
             else:
                 reader.set_exception(exc)
-            self._maybe_close_transport(fd)
-
-    def process_exited(self):
-        self._maybe_close_transport(-1)
         
-    def _maybe_close_transport(self, fd):
         if fd in self._pipe_fds:
             self._pipe_fds.remove(fd)
-        if len(self._pipe_fds) == 0:
+        self._maybe_close_transport(fd)
+
+    def process_exited(self):
+        self._process_exited = True
+        self._maybe_close_transport()
+        
+    def _maybe_close_transport(self):
+        if len(self._pipe_fds) == 0 and self._process_exited:
             self._transport.close()
             self._transport = None
 
--- a/asyncio/subprocess.py
+++ b/asyncio/subprocess.py
@@ -96,9 +96,9 @@ class SubprocessStreamProtocol(streams.FlowControlMixin,
     def _maybe_close_transport(self, fd):
         if fd in self._pipe_fds:
             self._pipe_fds.remove(fd)
-            if len(self._pipe_fds) == 0:
-                self._transport.close()
-                self._transport = None
+        if len(self._pipe_fds) == 0:
+            self._transport.close()
+            self._transport = None
 
 
 class Process:
--- a/asyncio/subprocess.py
+++ b/asyncio/subprocess.py
@@ -24,6 +24,7 @@ class SubprocessStreamProtocol(streams.FlowControlMixin,
         self._limit = limit
         self.stdin = self.stdout = self.stderr = None
         self._transport = None
+        self._pipe_fds = [-1]
 
     def __repr__(self):
         info = [self.__class__.__name__]
@@ -43,12 +44,14 @@ class SubprocessStreamProtocol(streams.FlowControlMixin,
             self.stdout = streams.StreamReader(limit=self._limit,
                                                loop=self._loop)
             self.stdout.set_transport(stdout_transport)
+            self._pipe_fds.append(1)
 
         stderr_transport = transport.get_pipe_transport(2)
         if stderr_transport is not None:
             self.stderr = streams.StreamReader(limit=self._limit,
                                                loop=self._loop)
             self.stderr.set_transport(stderr_transport)
+            self._pipe_fds.append(2)
 
         stdin_transport = transport.get_pipe_transport(0)
         if stdin_transport is not None:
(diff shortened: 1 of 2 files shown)


SethMichaelLarson:
As pretty much the sole author of `SubprocessStreamProtocol` I think you're the best to review this  @haypo. (If you don't want to be pinged in the future let me know! :) )

1st1:
cc @asvetlov. Although without a test there's nothing to review :)

SethMichaelLarson:
@1st1 Yes that's true, it'd be nice to get a nod to say this is actually a sane thing to do though. Doesn't break any other tests but I don't know what other untested assumptions I might be breaking! :P

1st1:
> Yes that's true, it'd be nice to get a nod to say this is actually a sane thing to do though

It looks sane. 

Asking for a review without tests isn't productive though, because you're essentially asking someone to stop doing what they are doing, understand (or remember, doesn't matter) your patch and **all code that it touches**, and make an opinion if your patch works. PRs without tests are usually reviewed the last if at all.

SethMichaelLarson:
Sorry, won't do it again. I added a simple test that fails on old build and passes with the proposed change. I presume it's okay to have a test that doesn't assert anything, just to make sure that no errors are thrown.

BotoX:
I can confirm that this pull request fixes my issue #484 
No more exceptions and no data is lost when the process exits, thanks! 👍

asyncio/subprocess.py
```diff
@@ -85,10 +88,17 @@ def pipe_connection_lost(self, fd, exc):
...
 
     def process_exited(self):
-        self._transport.close()
-        self._transport = None
+        self._maybe_close_transport(-1)
+        
+    def _maybe_close_transport(self, fd):
+        if fd in self._pipe_fds:
+            self._pipe_fds.remove(fd)
+            if len(self._pipe_fds) == 0:
```
1st1: Shouldn't this `if len(...)` be dedented?

sethmlarson: I guess that's true. My logic is we know that if nothing gets removed it won't be closed.

asyncio/subprocess.py
```diff
@@ -24,6 +24,7 @@ def __init__(self, limit, loop):
         self._limit = limit
         self.stdin = self.stdout = self.stderr = None
         self._transport = None
+        self._pipe_fds = [-1]
```
1st1: What does `-1` mean here?

SethMichaelLarson:
Added the change and CI passed. :)

asyncio/subprocess.py
sethmlarson: I should probably add a comment. `-1` is just the token for "process_exited" meaning we won't close a Transport until the process exits and all the pipes are done.

asyncio/subprocess.py
```diff
@@ -85,10 +88,17 @@ def pipe_connection_lost(self, fd, exc):
                 reader.feed_eof()
             else:
                 reader.set_exception(exc)
+            self._maybe_close_transport(fd)
```
1st1: I'd simply always call this method in `pipe_connection_lost`. Makes the code simpler to follow, and the logic with `if` statements should be (as they currently are) in `_maybe_close_transport`.

asyncio/subprocess.py
sethmlarson: Should I use a better sentinel object than -1?

asyncio/subprocess.py
sethmlarson: Sure, just dedent that line then?

asyncio/subprocess.py
1st1: Ah, interesting. I'd add a separate boolean attribute for that.

asyncio/subprocess.py
1st1: yep.

asyncio/subprocess.py
sethmlarson: Sure thing!

asyncio/subprocess.py
sethmlarson: Alright, just wanted to make sure I was understanding that comment correctly. :)

1st1:
Can you please close this PR and make a new one to http://github.com/python/cpython (with a link to this discussion and review). It would make merging this way easier for me.

sethmlarson:
So I made another change to `_maybe_close_transport()` to accommodate the flag for process exit.
Removed the fd parameter (since process exit isn't -1 anymore) and moved the removal into the `pipe_connection_lost`.

sethmlarson:
@1st1 Oh, sure! Can I do this when I'm home from work?  Don't have my actual environment all setup here, making do with the GitHub web interface. :)

1st1:
I like it now.  Thanks for opening the new PR! I'll close this one.

PR #497, opened 2017-02-22
Fix asyncio to support instantiation of new event loops in subprocesses

1st1:
See PR #452 for context.


1st1:
3.6.1 is scheduled to be released very soon so this PR can't wait. 

@gvanrossum @asvetlov @fafhrd91 please take a look at the commit in case I messed something up.

fafhrd91:
lgtm

